PR #13, opened 2026-08-09
Add document deletion from UI

### LINKED ISSUE #8: Document deletion from UI
missing feature )

though its available from pocketbase admin ui.

buldezir:
## Summary
- Add a Delete button on the document detail page with confirm, PocketBase delete, and navigate home
- Cover the flow with a Playwright e2e test

Fixes #8

## Test plan
- [x] `./scripts/test-all.sh` (unit + API e2e + Playwright)
- [ ] Manually: open a document → Delete → confirm → land on Documents list without that item
- [ ] Manually: open a document → Delete → cancel → stay on detail page

### DIFF
--- a/frontend/e2e/documents.spec.ts
+++ b/frontend/e2e/documents.spec.ts
@@ -47,7 +47,23 @@ test('delete document from detail page', async ({ page }) => {
 
   await expect(page).toHaveURL(/\/$/)
   await expect(page.getByRole('heading', { name: 'Documents' })).toBeVisible()
+  await expect(page.getByText('Loading documents...')).toHaveCount(0)
   await expect(page.locator(`main a[href="/document/${documentId}"]`)).toHaveCount(0)
+})
+
+test('cancel document delete keeps detail page', async ({ page }) => {
+  await loginAsUser(page)
+  await uploadFixture(page, 'sample.txt')
+
+  const documentURL = page.url()
+  const documentId = documentURL.match(/\/document\/([^/?#]+)/)?.[1]
+  expect(documentId).toBeTruthy()
+
+  page.once('dialog', (dialog) => dialog.dismiss())
+  await page.getByRole('button', { name: 'Delete' }).click()
+
+  await expect(page).toHaveURL(new RegExp(`/document/${documentId}`))
+  await expect(page.getByRole('button', { name: 'Delete' })).toBeEnabled()
 })
 
 test('reject duplicate file upload', async ({ page }) => {
--- a/frontend/e2e/documents.spec.ts
+++ b/frontend/e2e/documents.spec.ts
@@ -34,6 +34,22 @@ test('edit document title on detail page', async ({ page }) => {
   await expect(page.getByRole('heading', { name: 'E2E Edited Title' })).toBeVisible()
 })
 
+test('delete document from detail page', async ({ page }) => {
+  await loginAsUser(page)
+  await uploadFixture(page, 'sample.txt')
+
+  const documentURL = page.url()
+  const documentId = documentURL.match(/\/document\/([^/?#]+)/)?.[1]
+  expect(documentId).toBeTruthy()
+
+  page.once('dialog', (dialog) => dialog.accept())
+  await page.getByRole('button', { name: 'Delete' }).click()
+
+  await expect(page).toHaveURL(/\/$/)
+  await expect(page.getByRole('heading', { name: 'Documents' })).toBeVisible()
+  await expect(page.locator(`main a[href="/document/${documentId}"]`)).toHaveCount(0)
+})
+
 test('reject duplicate file upload', async ({ page }) => {
   await loginAsUser(page)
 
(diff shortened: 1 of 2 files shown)


frontend/e2e/documents.spec.ts
```diff
@@ -34,6 +34,22 @@ test('edit document title on detail page', async ({ page }) => {
...
+  const documentURL = page.url()
+  const documentId = documentURL.match(/\/document\/([^/?#]+)/)?.[1]
+  expect(documentId).toBeTruthy()
+
+  page.once('dialog', (dialog) => dialog.accept())
+  await page.getByRole('button', { name: 'Delete' }).click()
+
+  await expect(page).toHaveURL(/\/$/)
+  await expect(page.getByRole('heading', { name: 'Documents' })).toBeVisible()
+  await expect(page.locator(`main a[href="/document/${documentId}"]`)).toHaveCount(0)
```
buldezir, line 51: This assertion can pass even if the document was never deleted. `IndexPage` renders `Loading documents...` and no `a[href="/document/..."]` links at all while the list request is in flight, so `toHaveCount(0)` is satisfied immediately on arrival at `/` and the test never actually observes the loaded list.

Wait for the list to settle before asserting absence, e.g.:

```ts
await expect(page.getByText('Loading documents...')).toHaveCount(0)
await expect(page.locator(`main a[href="/document/${documentId}"]`)).toHaveCount(0)
```

Or assert on a positive signal first (another document link visible, or the `No documents yet.` empty state) so the negative assertion is meaningful.

frontend/src/routes/document.$documentId.tsx
```diff
@@ -169,6 +171,31 @@ export function DocumentDetailPage() {
...
+    if (!confirmed) {
+      return
+    }
+
+    try {
+      setDeleting(true)
+      setMessage('')
+      setError('')
+      await pb.collection('documents').delete(document.id)
+      await navigate({ to: '/' })
```
buldezir: There is a race with the realtime subscription set up in the mount effect. `pb.collection('documents').subscribe(documentId, () => load())` ignores the event action, so the `delete` event PocketBase emits for this record triggers a fresh `load()`. If that event lands before `navigate({ to: '/' })` unmounts the page, `getOne` 404s and the component flips to `Loading document...` and then a red `Failed to load document` message, which is a confusing flash on a successful delete.

Easiest fix is to ignore delete events in the subscription callback:

```ts
void pb.collection('documents').subscribe(documentId, (event) => {
  if (event.action === 'delete') {
    return
  }
  load()
})
```

Related but smaller: `navigate` is inside the `try`, so if it ever rejects the user sees `Failed to delete document` for a document that was in fact deleted. Moving the `navigate` call after the `try/catch` (or narrowing the `try` to the `delete`) avoids the misleading message.

frontend/e2e/documents.spec.ts
```diff
@@ -34,6 +34,22 @@ test('edit document title on detail page', async ({ page }) => {
...
+test('delete document from detail page', async ({ page }) => {
+  await loginAsUser(page)
+  await uploadFixture(page, 'sample.txt')
+
+  const documentURL = page.url()
+  const documentId = documentURL.match(/\/document\/([^/?#]+)/)?.[1]
+  expect(documentId).toBeTruthy()
+
+  page.once('dialog', (dialog) => dialog.accept())
+  await page.getByRole('button', { name: 'Delete' }).click()
```
buldezir, line 46: The PR test plan lists the cancel path as a manual-only check, but it is a real code branch (`if (!confirmed) return`) and cheap to cover here:

```ts
page.once('dialog', (dialog) => dialog.dismiss())
await page.getByRole('button', { name: 'Delete' }).click()
await expect(page).toHaveURL(new RegExp(`/document/${documentId}`))
await expect(page.getByRole('button', { name: 'Delete' })).toBeEnabled()
```

Without it, a regression that deletes without confirming (or disables the button on cancel) passes CI.

frontend/src/routes/document.$documentId.tsx
```diff
@@ -329,6 +356,14 @@ export function DocumentDetailPage() {
               Open file
             </a>
           )}
+          <button
+            type="button"
+            onClick={() => void onDelete()}
+            disabled={deleting}
```
buldezir, line 367: Delete is enabled regardless of `processing_status`, while reprocess on the same page is guarded by `canReprocess` (excludes `pending`/`processing`). Deleting a document whose pipeline is still running means `PipelineRunner.Run` keeps going against records that no longer exist: the `OnRecordDelete` hook removes the `processing_jobs` rows, and the runner's subsequent `Save(job)` / `Save(document)` calls fail and get logged as errors.

Was allowing this intentional? If so it would be good to say so, since the new e2e test exercises exactly that path (`uploadFixture` only waits for the `/document/` URL, not for processing to finish, so the delete happens mid-pipeline). Otherwise, either disable the button for `pending`/`processing` like reprocess does, or mention in the confirm text that processing will be aborted.

buldezir (commented):
Reviewed the delete flow. The core implementation looks right (owner-scoped `DeleteRule` exists, and the backend `OnRecordDelete("documents")` hook already cleans up `processing_jobs`), so nothing here is a correctness blocker. Four things worth addressing: one test assertion that can pass even when the delete fails, a realtime-subscription race on the detail page, missing coverage for the cancel branch, and delete being allowed mid-processing.

PR #14, opened 2026-08-09
Import from Paperless-ngx (#12)

### LINKED ISSUE #12: import from Paperless-ngx
Any chance for a Paperless-ngx import using endpoint URL and API key?

buldezir:
## Summary
- Adds one-shot Paperless-ngx import (URL + API key) that upserts tags, correspondents, and document types, then downloads documents with remote OCR text.
- Imported documents go through the normal pipeline so OCR is skipped when text is present and AI metadata extraction still runs.
- Admin UI at `/import` (More menu) plus `POST /api/app/import/ngx`; API key is not persisted; exact checksum duplicates are skipped.

## Test plan
- [ ] Manually import from a real Paperless-ngx instance via `/import`
- [ ] Confirm re-import of the same files reports skipped duplicates
- [ ] Confirm non-admin users do not see Import in the menu

Fixes #12

### DIFF
--- a/backend/e2e/import_ngx_test.go
+++ b/backend/e2e/import_ngx_test.go
@@ -1,0 +1,159 @@
+package e2e
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestImportNgxForbiddenForUser(t *testing.T) {
+	h := StartShared(t)
+	token := h.userToken(t)
+	status, raw := h.doJSON(t, http.MethodPost, "/api/app/import/ngx", token, map[string]any{
+		"url":     "http://example.com",
+		"api_key": "x",
+	})
+	if status != http.StatusForbidden {
+		t.Fatalf("status=%d body=%s", status, raw)
+	}
+}
+
+func TestImportNgxValidation(t *testing.T) {
+	h := StartShared(t)
+	token := h.adminUserToken(t)
+	status, raw := h.doJSON(t, http.MethodPost, "/api/app/import/ngx", token, map[string]any{
+		"url":     "",
+		"api_key": "x",
+	})
+	if status != http.StatusBadRequest {
+		t.Fatalf("status=%d body=%s", status, raw)
+	}
+	status, raw = h.doJSON(t, http.MethodPost, "/api/app/import/ngx", token, map[string]any{
+		"url":     "http://example.com",
+		"api_key": "",
+	})
+	if status != http.StatusBadRequest {
+		t.Fatalf("status=%d body=%s", status, raw)
+	}
+}
+
+func TestImportNgxFromRemote(t *testing.T) {
+	h := StartShared(t)
+
+	corrID := 2
+	typeID := 3
+	payload := []byte("imported-file-body-" + t.Name())
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/api/tags/", func(w http.ResponseWriter, r *http.Request) {
+		assertToken(t, r, "remote-token")
+		writeNgxPage(w, []map[string]any{{"id": 1, "name": "ImportedTag"}})
+	})
+	mux.HandleFunc("/api/correspondents/", func(w http.ResponseWriter, r *http.Request) {
+		writeNgxPage(w, []map[string]any{{"id": corrID, "name": "ImportedCorp"}})
+	})
+	mux.HandleFunc("/api/document_types/", func(w http.ResponseWriter, r *http.Request) {
+		writeNgxPage(w, []map[string]any{{"id": typeID, "name": "ImportedType"}})
+	})
+	mux.HandleFunc("/api/documents/", func(w http.ResponseWriter, r *http.Request) {
+		if strings.Contains(r.URL.Path, "/download") {
+			w.Header().Set("Content-Disposition", `attachment; filename="note.txt"`)
+			w.Header().Set("Content-Type", "text/plain")
+			_, _ = w.Write(payload)
+			return
+		}
+		writeNgxPage(w, []map[string]any{{
+			"id":                 42,
+			"title":              "RemoteInvoice",
+			"content":            "Imported OCR content for AI",
+			"tags":               []int{1},
+			"correspondent":      corrID,
+			"document_type":      typeID,
+			"created_date":       "2024-06-15",
+			"original_file_name": "note.txt",
+		}})
+	})
+	remote := httptest.NewServer(mux)
+	t.Cleanup(remote.Close)
+
+	token := h.adminUserToken(t)
+	status, raw := h.doJSON(t, http.MethodPost, "/api/app/import/ngx", token, map[string]any{
+		"url":     remote.URL,
+		"api_key": "remote-token",
+	})
+	requireStatus(t, status, http.StatusOK, raw)
+
+	var result map[string]any
+	if err := json.Unmarshal([]byte(raw), &result); err != nil {
+		t.Fatalf("decode: %v body %s", err, raw)
+	}
+	if intFromAny(result["imported"]) != 1 {
+		t.Fatalf("imported=%v body=%s", result["imported"], raw)
+	}
+	if intFromAny(result["tags_upserted"]) != 1 {
+		t.Fatalf("tags_upserted=%v", result["tags_upserted"])
+	}
+	if intFromAny(result["failed"]) != 0 {
+		t.Fatalf("failed=%v errors=%v", result["failed"], result["errors"])
+	}
+
+	// Second import of same bytes should skip as duplicate.
+	status, raw = h.doJSON(t, http.MethodPost, "/api/app/import/ngx", token, map[string]any{
+		"url":     remote.URL,
+		"api_key": "remote-token",
+	})
+	requireStatus(t, status, http.StatusOK, raw)
+	if err := json.Unmarshal([]byte(raw), &result); err != nil {
+		t.Fatalf("decode: %v", err)
+	}
+	if intFromAny(result["imported"]) != 0 || intFromAny(result["skipped_duplicates"]) != 1 {
+		t.Fatalf("second import result=%s", raw)
+	}
+
+	docsStatus, docsRaw := h.doJSON(t, http.MethodGet,
+		`/api/collections/documents/records?filter=title~"RemoteInvoice"&perPage=50`, token, nil)
+	requireStatus(t, docsStatus, http.StatusOK, docsRaw)
+	var docs struct {
+		Items []map[string]any `json:"items"`
+	}
+	if err := json.Unmarshal([]byte(docsRaw), &docs); err != nil {
+		t.Fatalf("decode docs: %v", err)
+	}
+	if len(docs.Items) != 1 {
+		t.Fatalf("expected 1 document, got %d: %s", len(docs.Items), docsRaw)
+	}
+	ocr, _ := docs.Items[0]["ocr_text"].(string)
+	if ocr != "Imported OCR content for AI" {
+		t.Fatalf("ocr_text=%q", ocr)
+	}
+}
+
+func assertToken(t *testing.T, r *http.Request, want string) {
+	t.Helper()
+	got := r.Header.Get("Authorization")
+	if got != "Token "+want {
+		t.Fatalf("Authorization=%q", got)
+	}
+}
+
+func writeNgxPage(w http.ResponseWriter, results []map[string]any) {
+	_ = json.NewEncoder(w).Encode(map[string]any{
+		"count":    len(results),
+		"next":     nil,
+		"previous": nil,
+		"results":  results,
+	})
+}
+
+func intFromAny(v any) int {
+	switch n := v.(type) {
+	case float64:
+		return int(n)
+	case int:
+		return n
+	default:
+		return -1
+	}
+}
--- a/backend/e2e/import_ngx_test.go
+++ b/backend/e2e/import_ngx_test.go
@@ -1,11 +1,14 @@
 package e2e
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
+	"time"
 )
 
 func TestImportNgxForbiddenForUser(t *testing.T) {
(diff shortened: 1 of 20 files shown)


backend/internal/ngximport/import.go
```diff
@@ -0,0 +1,317 @@
...
+		tagIDs := make([]string, 0, len(doc.Tags))
+		for _, ngxTagID := range doc.Tags {
+			if id := tagMap[ngxTagID]; id != "" {
+				tagIDs = append(tagIDs, id)
+			}
+		}
+		if len(tagIDs) > 0 {
+			record.Set("tags", tagIDs)
+		}
+	}
```
buldezir: These assignments do not survive processing. The `OnRecordCreate("documents")` hook queues a full-pipeline job for every saved record, and `ApplyMetadataStep` later does `state.Document.Set("tags", tagIDs)` with only the AI-derived tags, while `applyCorrespondent`/`applyDocumentType` re-set those fields (and set them to `""` when the AI returns nothing). `applyExtractedMetadata` likewise overwrites `title`.

So the imported ngx taxonomy is written and then replaced a few seconds later, which is probably not what a user migrating a curated library expects. Options: merge imported tags with AI tags in `ApplyMetadataStep`, or import with a reduced step list (skip `extract_metadata`/`apply_metadata` when the remote already provided title/correspondent/type), or at minimum document that AI output wins.

backend/internal/ngximport/client.go
```diff
@@ -0,0 +1,240 @@
...
+	req, err := c.newRequest(http.MethodGet, rel, nil)
+	if err != nil {
+		return downloadedFile{}, err
+	}
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return downloadedFile{}, err
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(io.LimitReader(resp.Body, 32<<20))
```
buldezir: A document larger than 32 MiB is silently truncated here rather than failing: `io.ReadAll(io.LimitReader(...))` returns the first 32 MiB with a nil error, and the caller then hashes and stores that partial body as a complete document. The corruption is invisible, and the checksum makes it look legitimate.

Reading one extra byte and erroring when the limit is hit makes the failure explicit:

```go
const maxDownloadBytes = 32 << 20
body, err := io.ReadAll(io.LimitReader(resp.Body, maxDownloadBytes+1))
if err != nil {
	return downloadedFile{}, err
}
if len(body) > maxDownloadBytes {
	return downloadedFile{}, fmt.Errorf("document %d exceeds %d bytes", id, maxDownloadBytes)
}
```

Related: the default client's 60s timeout covers the whole request including the body transfer, so large documents over a slow link will fail. Consider a longer timeout for downloads than for list calls.

backend/internal/ngximport/client.go
```diff
@@ -0,0 +1,240 @@
...
+	req, err := c.newRequest(http.MethodGet, relOrPath, nil)
+	if err != nil {
+		return err
+	}
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(io.LimitReader(resp.Body, 8<<20))
```
buldezir: The 8 MiB cap is likely to be hit in practice. The ngx `/api/documents/` list response embeds the full OCR `content` for every result, and `defaultPageSize` is 100, so 100 documents averaging 80 KB of text already exceeds it. When that happens the truncated body reaches `json.Unmarshal` and the admin sees `decode /api/documents/...: unexpected end of JSON input`, with no hint that a size limit was the cause.

Streaming with `json.NewDecoder(resp.Body).Decode(dest)` avoids the cap for the success path (read the body only for the error branch), or lower `defaultPageSize` for documents and raise the limit.

backend/internal/ngximport/client.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (c *Client) ListCorrespondents() ([]namedEntity, error) {
+	return listAll[namedEntity](c, "/api/correspondents/")
+}
+
+func (c *Client) ListDocumentTypes() ([]namedEntity, error) {
+	return listAll[namedEntity](c, "/api/document_types/")
+}
+
+func (c *Client) ListDocuments() ([]ngxDocument, error) {
```
buldezir, line 124: `listAll` accumulates every document, including full OCR text, in memory before the import loop starts. For a library of tens of thousands of documents that is a large allocation held for the entire run, and nothing is imported until the last page has been fetched.

A callback-based variant (`forEachPage(func([]ngxDocument) error)`) would let `Run` import page by page and would also give the user partial progress if the run is interrupted.

backend/internal/ngximport/import.go
```diff
@@ -0,0 +1,317 @@
...
+			0,
+			map[string]any{"name": name},
+		)
+		if findErr == nil && len(existing) > 0 {
+			return existing[0].Id, nil
+		}
+		return "", err
+	}
+	return record.Id, nil
+}
```
buldezir: This re-implements `worker.ensureNamedEntity` with different matching rules, which will create duplicate entities. The worker looks up `name_original` first and only then `name`; this version matches on `name` only.

Concretely: the AI pipeline created a correspondent with `name = "Acme"` and `name_original = "ACME GMBH"` (translation is applied when `processing_result_language` is set). The remote ngx instance has a correspondent named `ACME GMBH`. This lookup finds nothing, so a second record is created with both fields set to `ACME GMBH`, and the user ends up with two correspondents for the same company.

Moving `ensureNamedEntity`/`findNamedEntity` into a shared package and calling it from both places would keep the matching rules in one place.

backend/internal/ngximport/import.go
```diff
@@ -0,0 +1,317 @@
...
+		if errors.As(err, &dup) {
+			return dup
+		}
+		if dup := duplicates.ErrDuplicateFromSaveConflict(app, record, err); dup != nil {
+			return dup
+		}
+		if strings.Contains(err.Error(), "document already exists") {
+			return &duplicates.ErrDuplicate{}
+		}
+		return err
```
buldezir, line 286: The duplicate detection here is fragile in both branches. `errors.As(err, &dup)` cannot match: the create hook converts the duplicate into `router.NewBadRequestError(...)`, never an `*ErrDuplicate`. So the substring match on `"document already exists"` is the branch that actually fires, and it depends on the exact wording of `ErrDuplicate.Error()`. If that message is ever reworded, imports start reporting duplicates as failures instead.

The hook already attaches the structured `{"duplicate_of": id}` payload to the `ApiError`; inspecting that (or adding a helper in `duplicates` that both call) would be more durable and would preserve `ExistingID`, which the current fallback drops.

backend/internal/ngximport/import.go
```diff
@@ -0,0 +1,317 @@
...
+			continue
+		}
+		localID, err := ensureNamed(app, collection, name)
+		if err != nil {
+			return nil, upserted, err
+		}
+		idMap[entity.ID] = localID
+		upserted++
+	}
+	return idMap, upserted, nil
```
buldezir: `upserted` counts every entity seen, including ones that already existed locally, so the UI's "Tags upserted: N" reports the size of the remote taxonomy rather than what changed. Re-running an import shows the same number every time, which makes the summary hard to interpret.

Having `ensureNamed` report whether it created the record and counting only creations would make the number mean what the label says.

backend/internal/appapi/import_ngx.go
```diff
@@ -0,0 +1,67 @@
...
+			return writeError(e, http.StatusBadRequest, err.Error())
+		}
+
+		result, err := ngximport.Run(app, ownerID, req.URL, req.APIKey)
+		if errors.Is(err, ngximport.ErrImportInProgress) {
+			return writeError(e, http.StatusConflict, "An import is already in progress.")
+		}
+		if err != nil {
+			return writeError(e, http.StatusBadRequest, "Import failed: "+err.Error())
+		}
```
buldezir, line 48: Two things about this error path:

Every failure maps to 400, including ones the caller cannot fix: an unreachable host, a 500 from the remote instance, or a local DB error all come back as "Bad Request". `handlePostDuplicatesScan` uses 500 for its equivalent case; 502 would fit the remote failures.

More importantly, `ngximport.Run` performs the entire migration inline, so this request stays open for the whole run. Against a real library that will hit reverse-proxy and browser timeouts long before it finishes, and because `importBusy` is still held, the retry then fails with 409 while the original run continues invisibly. A job record plus a polled status endpoint (the pattern `processing_jobs` already establishes) would make this usable for anything beyond a small instance.

backend/internal/appapi/appapi.go
```diff
@@ -23,6 +23,7 @@ func Register(app core.App, rt *config.Runtime) {
 			g.GET("/settings", bindAdmin(handleGetSettings(app, rt)))
 			g.PATCH("/settings", bindAdmin(handlePatchSettings(app, rt)))
 			g.POST("/duplicates/scan", bindAdmin(handlePostDuplicatesScan(app, rt)))
+			g.POST("/import/ngx", handlePostImportNgx(app))
```
buldezir: This is the only route in the group that applies `bindAdmin` inside the handler rather than at registration. It works, but reading this list it is the one line that looks like it forgot an auth guard. Returning the plain handler and wrapping here, as the neighbouring routes do, keeps the guard visible where the route is declared.

buldezir (commented):
## Review

Solid, well-scoped feature: the client is clean, `NormalizeBaseURL` handles the `/api` suffix cases, taxonomy mapping is done up front, and reusing the normal document-create path so OCR is skipped when `ocr_text` is present is the right call. Docs, unit tests, API e2e, and a Playwright spec are all present. `go test ./internal/ngximport/...` passes locally.

The findings below are mostly about behavior on a real (large) instance rather than the happy path exercised by the tests.

**Main concerns**

1. Per-document tags, correspondent, and document type set at import time are overwritten by the AI pipeline. `ApplyMetadataStep` unconditionally does `document.Set("tags", tagIDs)` and `applyCorrespondent`/`applyDocumentType` re-set (or clear) those fields from the AI result, and `applyExtractedMetadata` overwrites `title`. Since every imported document is queued for the full pipeline, the mapping work in `importOneDocument` is discarded once the job runs.
2. Documents larger than 32 MB are silently truncated by the `io.LimitReader` in `DownloadDocument`, so a corrupt file is stored with a valid-looking checksum and no error.
3. The 8 MB `io.LimitReader` in `getJSON` is easy to exceed: the ngx document list includes the full OCR `content` per document, and the page size is 100.
4. `ensureNamed` duplicates `worker.ensureNamedEntity` but only matches on `name`, ignoring `name_original`, which will create duplicate correspondents/document types alongside AI-created ones.
5. The import runs synchronously inside the HTTP request, with no progress or resume. A library of a few thousand documents will exceed proxy/browser timeouts, and the client has no way to learn the outcome.

**Nits not worth inline comments**

- `backend/internal/ngximport/client.go`: the `ngxDocument` struct is not gofmt-clean (field alignment). `gofmt -w` fixes it.
- `RunWithClient` is exported and documented as "for tests", but no test uses it. Either drop it or use it to cover `Run`'s document loop without HTTP.
- Error messages echo up to 200 bytes of the remote response back to the caller, and the server will fetch any admin-supplied URL, including internal addresses. Admin-only makes this acceptable, but it is worth being aware of.

backend/e2e/import_ngx_test.go
```diff
@@ -0,0 +1,159 @@
...
+	requireStatus(t, docsStatus, http.StatusOK, docsRaw)
+	var docs struct {
+		Items []map[string]any `json:"items"`
+	}
+	if err := json.Unmarshal([]byte(docsRaw), &docs); err != nil {
+		t.Fatalf("decode docs: %v", err)
+	}
+	if len(docs.Items) != 1 {
+		t.Fatalf("expected 1 document, got %d: %s", len(docs.Items), docsRaw)
+	}
```
buldezir, line 125: This lookup races the pipeline. Creating the document enqueues a full-pipeline job that the shared harness can actually run (mock OCR and AI are seeded), and the mock AI returns title `Acme Plumbing Invoice INV-1001`. Once `apply_metadata` completes, the title is no longer `RemoteInvoice` and this filter returns zero items.

It passes today only because the job usually has not finished yet. Looking the document up by `checksum` (stable across processing) or capturing the id right after the import call would make this deterministic.

buldezir:
Addressed the metadata-overwrite review finding by adding an import mode choice:

- **`preserve`** (default): keep ngx title, tags, correspondent, document type, date, and OCR text; queue preview/OCR/duplicate detection only so AI does not overwrite metadata.
- **`reprocess`**: import original files only and run the full OCR + AI pipeline.

UI radios are on `/import`; API accepts `"mode": "preserve" | "reprocess"` on `POST /api/app/import/ngx`.

backend/internal/ngximport/client.go
buldezir: Fixed: downloads read max+1 bytes and error when the 32 MiB limit is exceeded. Download requests use a longer (5m) timeout than list/JSON calls.

backend/internal/ngximport/client.go
buldezir: Fixed: success path streams with json.NewDecoder; document pages use page_size=25.

backend/internal/ngximport/client.go
buldezir: Fixed: documents are imported via ForEachDocuments page-by-page instead of accumulating the full remote list.

backend/internal/ngximport/import.go
buldezir: Fixed: correspondents/document types go through worker.EnsureNamedEntity (name_original then name). Tags use worker.EnsureTag.

backend/internal/ngximport/import.go
buldezir: Fixed: Save failures unwrap ApiError duplicate_of via duplicates.ErrDuplicateFromAPIError; dropped the string-match fallback.

backend/internal/ngximport/import.go
buldezir: Fixed: *_upserted counts only newly created entities.

backend/internal/appapi/import_ngx.go
buldezir: Addressed with lightweight in-process async: POST returns 202 + job_id; GET /api/app/import/ngx/status?job_id=... polls. Job state is in-memory (not a DB collection); process restart loses it. Validation still 400; in-progress 409; run failures surface on the job.

backend/internal/appapi/appapi.go
buldezir: Fixed: bindAdmin wraps the routes at registration like the neighboring admin endpoints.

backend/e2e/import_ngx_test.go
buldezir: Fixed: e2e looks up the imported document by checksum and polls the async job to completion.

PR #16, opened 2026-08-12
Feature: download archive (#15)

### LINKED ISSUE #15: Feature: download archive
download archive of all user's documents.

options:
* only original files
* with ocr text (in .ocr.txt files with same name as original file)
* with all metadata (includes ocr option, + all extracted metadata in .metadata.json)

buldezir:
## Summary
- Adds `GET /api/app/documents/export?mode=originals|ocr|metadata` that streams a zip of the authenticated user's documents
- Adds an Export page (More menu) with mode radios for originals, OCR sidecars, and metadata JSON
- Zip layout: `{documentId}/{original}` plus optional `{stem}.ocr.txt` / `{stem}.metadata.json`

Fixes #15

## Test plan
- [ ] Manually open Export, download each mode, and confirm zip contents
- [ ] Confirm another user's documents do not appear in the archive

### DIFF
--- a/backend/internal/appapi/export.go
+++ b/backend/internal/appapi/export.go
@@ -43,6 +43,7 @@ func handleExportDocuments(app core.App) func(*core.RequestEvent) error {
 			fileKey := rec.BaseFilesPath() + "/" + name
 			docs = append(docs, ExportDocument{
 				ID:               rec.Id,
+				Title:            firstNonEmpty(rec.GetString("title"), rec.GetString("title_original"), "Untitled"),
 				OriginalFilename: name,
 				OpenFile: func() (io.ReadCloser, error) {
 					reader, err := fsys.GetReader(fileKey)
@@ -141,7 +142,6 @@ func buildExportMetadata(app core.App, record *core.Record, originalFilename str
 		"checksum":                record.GetString("checksum"),
 		"created":                 record.GetString("created"),
 		"updated":                 record.GetString("updated"),
-		"ocr_text":                record.GetString("ocr_text"),
 		"original_filename":       originalFilename,
 	}
 }
--- a/backend/internal/appapi/export.go
+++ b/backend/internal/appapi/export.go
@@ -1,0 +1,184 @@
+package appapi
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/pocketbase/dbx"
+	"github.com/pocketbase/pocketbase/core"
+)
+
+const exportPageSize = 100
+
+func handleExportDocuments(app core.App) func(*core.RequestEvent) error {
+	return func(e *core.RequestEvent) error {
+		mode, err := ParseExportMode(e.Request.URL.Query().Get("mode"))
+		if err != nil {
+			return writeError(e, http.StatusBadRequest, err.Error())
+		}
+
+		userID := e.Auth.Id
+		records, err := listOwnedDocuments(app, userID)
+		if err != nil {
+			return writeError(e, http.StatusInternalServerError, "Failed to list documents.")
+		}
+
+		fsys, err := app.NewFilesystem()
+		if err != nil {
+			return writeError(e, http.StatusInternalServerError, "Failed to open storage.")
+		}
+		defer fsys.Close()
+
+		docs := make([]ExportDocument, 0, len(records))
+		for _, record := range records {
+			fileName := record.GetString("file")
+			if fileName == "" {
+				continue
+			}
+			rec := record
+			name := fileName
+			fileKey := rec.BaseFilesPath() + "/" + name
+			docs = append(docs, ExportDocument{
+				ID:               rec.Id,
+				OriginalFilename: name,
+				OpenFile: func() (io.ReadCloser, error) {
+					reader, err := fsys.GetReader(fileKey)
+					if err != nil {
+						app.Logger().Warn("export skip missing file", "document", rec.Id, "error", err)
+						return nil, err
+					}
+					return reader, nil
+				},
+				OCRText:  rec.GetString("ocr_text"),
+				Metadata: buildExportMetadata(app, rec, name),
+			})
+		}
+
+		e.Response.Header().Set("Content-Type", "application/zip")
+		e.Response.Header().Set("Content-Disposition", `attachment; filename="paperless-export.zip"`)
+		e.Response.WriteHeader(http.StatusOK)
+
+		if err := WriteExportZip(e.Response, mode, docs); err != nil {
+			app.Logger().Error("export zip failed", "error", err)
+			return err
+		}
+		return nil
+	}
+}
+
+func listOwnedDocuments(app core.App, userID string) ([]*core.Record, error) {
+	var all []*core.Record
+	page := 1
+	for {
+		records, err := app.FindRecordsByFilter(
+			"documents",
+			"user = {:userId}",
+			"-created",
+			exportPageSize,
+			(page-1)*exportPageSize,
+			dbx.Params{"userId": userID},
+		)
+		if err != nil {
+			return nil, fmt.Errorf("list documents: %w", err)
+		}
+		all = append(all, records...)
+		if len(records) < exportPageSize {
+			break
+		}
+		page++
+	}
+	return all, nil
+}
+
+func buildExportMetadata(app core.App, record *core.Record, originalFilename string) map[string]any {
+	tags := make([]string, 0)
+	for _, tagID := range record.GetStringSlice("tags") {
+		if tagID == "" {
+			continue
+		}
+		tagRec, err := app.FindRecordById("tags", tagID)
+		if err != nil {
+			continue
+		}
+		if name := strings.TrimSpace(tagRec.GetString("name")); name != "" {
+			tags = append(tags, name)
+		}
+	}
+
+	documentType := ""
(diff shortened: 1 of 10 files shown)


backend/internal/appapi/export_zip.go
```diff
@@ -0,0 +1,170 @@
...
+// Entries are flattened under paperless-export/ as "[id] title.ext".
+// Documents that fail OpenFile or lack an ID/filename are skipped.
+// Empty OCR text omits the OCR sidecar.
+func WriteExportZip(w io.Writer, mode ExportMode, docs []ExportDocument) (err error) {
+	zw := zip.NewWriter(w)
+	defer func() {
+		if closeErr := zw.Close(); err == nil {
+			err = closeErr
+		}
+	}()
```
buldezir, line 65: The deferred `Close()` runs on the error path too, so a mid-stream failure still writes a valid central directory for whatever was packed so far. Since the handler has already sent `200` with the zip content type, the client ends up with a well-formed archive that is silently missing documents — `zip.NewReader` opens it happily. (I checked PocketBase's `router.ErrorHandler`: it sees the response as already written and returns early, so nothing is appended to the body either.)

Consider skipping `Close()` when `err != nil` so the download is detectably truncated:

```go
defer func() {
	if err != nil {
		return // leave the stream truncated so the client detects a bad archive
	}
	err = zw.Close()
}()
```

backend/internal/appapi/export_zip.go
```diff
@@ -0,0 +1,170 @@
...
+
+	for _, doc := range docs {
+		if doc.OpenFile == nil || strings.TrimSpace(doc.OriginalFilename) == "" || strings.TrimSpace(doc.ID) == "" {
+			continue
+		}
+
+		reader, openErr := doc.OpenFile()
+		if openErr != nil {
+			continue
+		}
```
buldezir, line 75: A document whose file is missing from storage is dropped from the archive with no signal to the user beyond a server-side `Warn`. This is the most likely way an export ends up incomplete in practice, and it looks identical to a successful export from the browser.

Worth writing a manifest entry into the zip (e.g. `paperless-export/export-report.txt` listing exported and skipped document IDs), which also gives users a way to verify the archive contents. Same idea would cover the truncation case above.

backend/internal/appapi/export_zip.go
```diff
@@ -0,0 +1,170 @@
...
+				continue
+			}
+			b.WriteByte(' ')
+			prevSpace = true
+		default:
+			b.WriteRune(r)
+			prevSpace = false
+		}
+	}
+	out := strings.TrimSpace(b.String())
```
buldezir, line 148: `sanitizeExportTitle` strips unsafe characters but does not cap length. Titles from `title_original` can be derived from long filenames, and `[id] <title>.metadata.json` can exceed the 255-byte per-component limit that most filesystems and extraction tools enforce, so those entries fail to extract.

Truncating the sanitized title (say 120 bytes, on a rune boundary) would make the entry names always extractable.

backend/internal/appapi/export.go
```diff
@@ -0,0 +1,184 @@
...
+				OpenFile: func() (io.ReadCloser, error) {
+					reader, err := fsys.GetReader(fileKey)
+					if err != nil {
+						app.Logger().Warn("export skip missing file", "document", rec.Id, "error", err)
+						return nil, err
+					}
+					return reader, nil
+				},
+				OCRText:  rec.GetString("ocr_text"),
+				Metadata: buildExportMetadata(app, rec, name),
```
buldezir: `buildExportMetadata` runs for every document even when `mode` is `originals` or `ocr`, where the result is thrown away. Each call issues one `FindRecordById` per tag plus one for `document_type` and one for `correspondent`, so exporting 2000 documents in `originals` mode costs several thousand pointless queries before the first byte is written.

Either guard it (`if mode == ExportModeMetadata`) or make it lazy like `OpenFile` (`Metadata func() map[string]any`). If you go lazy, a small per-request name cache for tags/types/correspondents would also cut the repeated lookups for shared entities in `metadata` mode.

backend/internal/appapi/export.go
```diff
@@ -0,0 +1,184 @@
...
+		mode, err := ParseExportMode(e.Request.URL.Query().Get("mode"))
+		if err != nil {
+			return writeError(e, http.StatusBadRequest, err.Error())
+		}
+
+		userID := e.Auth.Id
+		records, err := listOwnedDocuments(app, userID)
+		if err != nil {
+			return writeError(e, http.StatusInternalServerError, "Failed to list documents.")
+		}
```
buldezir, line 28: The underlying error is dropped here — the client gets a generic 500 and nothing reaches the log, unlike the zip failure at line 66 which does log. Add `app.Logger().Error("export list documents failed", "error", err)` (same for the `NewFilesystem` failure below) so these are diagnosable.

backend/internal/appapi/export.go
```diff
@@ -0,0 +1,184 @@
...
+			return nil, fmt.Errorf("list documents: %w", err)
+		}
+		all = append(all, records...)
+		if len(records) < exportPageSize {
+			break
+		}
+		page++
+	}
+	return all, nil
+}
```
buldezir, line 101: This materializes every document record — including `ocr_text`, which is the largest field — before a single byte is written. `WriteExportZip` is built to stream, so the handler could feed it page by page (pass an iterator or call it per page with a shared `zip.Writer`) and keep memory flat regardless of library size.

Minor related note: `rec := record` / `name := fileName` at lines 41-42 are leftovers from pre-1.22 loop variable capture and are no longer needed on `go 1.26.3`.

frontend/src/lib/pocketbase.ts
```diff
@@ -628,6 +628,46 @@ export async function importFromNgx(url: string, apiKey: string, mode: NgxImport
...
+    const anchor = document.createElement('a')
+    anchor.href = objectUrl
+    anchor.download = 'paperless-export.zip'
+    anchor.rel = 'noopener'
+    document.body.appendChild(anchor)
+    anchor.click()
+    anchor.remove()
+  } finally {
+    URL.revokeObjectURL(objectUrl)
+  }
```
buldezir, line 668: Two things here:

1. `response.blob()` buffers the entire archive in tab memory before the save dialog appears, so a multi-GB export can OOM the tab despite the server streaming it. There is no clean fix while the endpoint requires an `Authorization` header, but a short-lived signed download URL (navigable directly) would let the browser stream to disk. Worth at least a comment noting the limitation.
2. `URL.revokeObjectURL` in the `finally` block runs synchronously right after `anchor.click()`. Chromium tolerates this, but Safari and older Firefox can abort the download when the URL is revoked in the same task. Deferring it (`setTimeout(() => URL.revokeObjectURL(objectUrl), 0)`, or on `window` unload) is the safer pattern.

frontend/src/routes/export.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  async function onSubmit(event: SubmitEvent<HTMLFormElement>) {
+    event.preventDefault()
+    try {
+      setRunning(true)
+      setError('')
+      setSuccess('')
+      await downloadDocumentsArchive(mode)
+      setSuccess('Archive download started.')
+    } catch (err) {
```
buldezir, line 66: By the time this runs the whole archive has already been fetched and handed to the browser, so "Archive download started." understates what happened — and it is shown even when the server skipped documents. Something like "Archive ready — check your downloads." reads more accurately, and if you add the manifest suggested in `export_zip.go`, mentioning it here would make partial exports discoverable.

Also, since `Preparing archive…` can sit there for a long time on a big library with no progress indication, consider showing the document count on this page so users know what to expect.

buldezir (commented):
## Review

Solid, well-tested feature: the zip writer is cleanly separated from the handler and covered by unit tests, and the e2e tests cover mode selection, invalid mode, unauthenticated access, and — importantly — owner isolation. I verified locally at `b6b059d`: `go vet ./internal/appapi ./e2e`, `go test ./internal/appapi`, `tsc -b`, and eslint on the changed frontend files all pass.

No blocking bugs found. The main themes worth addressing before merge:

1. **Partial archives are indistinguishable from complete ones.** Two paths silently drop documents while still returning `200` and a structurally valid zip: a mid-stream write error (the deferred `zw.Close()` still writes a valid central directory) and any document whose file is missing from storage. The user sees "Archive download started." either way. Details inline.
2. **Metadata is built for every document regardless of mode.** `buildExportMetadata` issues one query per tag plus one each for document type and correspondent, and in `originals`/`ocr` mode the result is discarded. For a large library this is thousands of wasted queries per export.
3. **Everything is buffered before it is streamed** — all records server-side, and the entire zip in browser memory client-side. Fine for typical libraries, but it undercuts the streaming design of `WriteExportZip`.

Smaller notes on filename length, error logging, and object URL lifetime are inline.

Two things outside the diff:

- The `Features` list in `README.md` and the endpoint documentation in `docs/development.md` (which documents `POST /api/app/import/ngx` in similar detail) were not updated for the new Export page and `GET /api/app/documents/export`.
- `bindAuth` also accepts a `_superusers` token, in which case `e.Auth.Id` is a superuser record ID and the filter `user = {:userId}` matches nothing, so a superuser gets an empty zip with `200`. Probably acceptable, but worth a deliberate decision (e.g. `400`/explicit message) rather than an empty archive.

PR #18, opened 2026-08-12
Add named AI providers and per-task model selection

### LINKED ISSUE #17: OpenRouter Support
It would be great if instead of needing a specific Mistral or google OCR API, there was an integration with OpenRouter so the user could choose whatever models they wanted for each task.

buldezir:
## Summary
- Admins can add named LLM/OCR providers (`openai`, `openrouter`, `mistral`, `google_vision`) with their own base URL and API key, then bind OCR, extraction, chat, and search to **provider → model** instead of one shared OpenAI endpoint.
- OpenAI/OpenRouter can run OCR via chat completions (images as `image_url`, PDFs as file parts). OpenRouter OCR model lists are filtered with `input_modalities=file`.
- First-boot env values and existing singleton keys migrate into `ai_providers`; old settings columns are left in place.

Fixes #17

## Test plan
- [ ] Docker image build (daemon unavailable locally)
- [ ] Settings: add/edit/delete a provider; confirm in-use providers cannot be deleted
- [ ] Settings: pick OCR/extract/chat/search provider+model; save reloads runtime
- [ ] Setup wizard: add a provider, then assign OCR + extraction models
- [ ] OCR test page: choose a provider from the catalog and extract text
- [ ] Upload a PDF/image through the normal pipeline with an OpenRouter/OpenAI OCR provider


backend/internal/worker/processor.go
```diff
@@ -147,39 +147,39 @@ func createProcessingJob(app core.App, documentID string, steps []string, forceS
...
+		}
+		if err := p.runJob(jobs[0].Id); err != nil {
+			p.app.Logger().Error("job error", "job", jobs[0].Id, slog.Any("error", err))
+		}
 	}
+}
 
-	return p.runJob(jobs[0].Id)
+func (p *Processor) processNextPending() error {
+	p.drainPending()
```
buldezir, line 183: This changes cron behavior in a way that's easy to miss, and it's unrelated to the provider/model work in this PR.

Previously `processNextPending` used `TryLock` and returned immediately when a job was already running. Now it calls `drainPending`, which does a blocking `p.processing.Lock()`. The cron is registered as:

```go
app.Cron().MustAdd("process_pending_jobs", cronExpr, func() {
	if err := p.processNextPending(); err != nil { ... }
})
```

with a default expression of `* * * * *`. So while a job runs for, say, 8 minutes, 8 cron goroutines pile up blocked on the mutex, then wake in sequence and each do a redundant drain pass (all but the first find no pending rows). It's bounded by job duration rather than unbounded, but it's still goroutine accumulation that didn't exist before, and it defeats the point of the cron being a cheap periodic poke.

Keeping the `TryLock` fast path preserves the old semantics while still getting the drain loop:

```go
func (p *Processor) drainPending() {
	if !p.processing.TryLock() {
		return
	}
	defer p.processing.Unlock()
	// ... loop unchanged
}
```

`dispatch` can call this directly too — a dropped `dispatch` is harmless now that the winner drains everything pending rather than a single job.

Separately: the drain loop itself (and the `document.$documentId.tsx` polling change) looks like a real fix for jobs stalling when a second upload lands mid-processing, but neither is mentioned in the PR description. Either call it out in the description or split it into its own PR — it's the kind of change that deserves to be findable in history on its own.

backend/internal/appapi/appapi.go
```diff
@@ -19,10 +19,15 @@ func Register(app core.App, rt *config.Runtime) {
...
-			g.GET("/ocr/providers", bindAuth(handleOCRProviders(rt)))
+			g.GET("/ocr/providers", bindAuth(handleOCRProviders(app, rt)))
 			g.POST("/ocr/test", bindAuth(handleOCRTest(app, rt)))
 			g.GET("/settings", bindAdmin(handleGetSettings(app, rt)))
 			g.PATCH("/settings", bindAdmin(handlePatchSettings(app, rt)))
+			g.GET("/providers", bindAdmin(handleListProviders(app)))
+			g.POST("/providers", bindAdmin(handleCreateProvider(app, rt)))
+			g.PATCH("/providers/{id}", bindAdmin(handlePatchProvider(app, rt)))
+			g.DELETE("/providers/{id}", bindAdmin(handleDeleteProvider(app, rt)))
+			g.GET("/providers/{id}/models", bindAuth(handleListProviderModels(app)))
```
buldezir: This should be `bindAdmin`. Every other `/providers` route on the lines above is admin-gated, and this one exposes the same resource.

Two concrete consequences for a regular authenticated user:

1. They can make the server issue arbitrary requests to the configured provider using the admin's API key, as often as they like. `ListModels` has a 20s timeout and no rate limiting, so it's a free outbound-request primitive.
2. On failure the handler returns `err.Error()` verbatim (see the comment on `providers.go`), which includes up to 300 characters of the upstream response body.

They can already discover provider IDs because `GET /ocr/providers` is `bindAuth`, so no extra guessing is needed.

If the OCR test page needs the catalog for non-admins, it'd be better to scope that to the providers actually bound for OCR rather than opening the whole catalog endpoint.

buldezir (commented):
## Review

The provider/model refactor is well structured: `aiprovider` is a clean seam, the legacy migration path is thought through (env seeding, singleton migration, legacy columns kept), and the runtime now degrades gracefully when a task has no usable provider instead of constructing a dead client. Test coverage tracks the change at every layer (unit, API e2e, Playwright), and I verified locally that `go build`, `go vet`, the backend unit tests, and the frontend build all pass on this branch.

The findings below are mostly about the edges of the new admin surface rather than the core design. The two I'd want addressed before merge are the worker concurrency change (which is unrelated to this PR and is a behavioral regression) and the `bindAuth` on the model-catalog endpoint. The rest are correctness/UX papercuts in the new Settings flow.

### Blocking

1. **`processNextPending` now blocks on the mutex** (`backend/internal/worker/processor.go`). The cron tick used to bail out immediately via `TryLock`; it now waits for the running job to finish, so cron goroutines accumulate for the duration of a long job. This is also unrelated to provider/model selection and isn't mentioned in the PR description.
2. **`GET /providers/{id}/models` is `bindAuth`, not `bindAdmin`** (`backend/internal/appapi/appapi.go`). Any authenticated non-admin can drive outbound requests with the admin's API key, and the handler returns the raw upstream error body back to them.

### Worth fixing

3. `handlePatchProvider` lets an admin change `sdk` on a provider that extraction/chat/search are bound to, silently disabling the LLM and re-gating the whole app behind the setup wizard.
4. `lookupProvider` collapses every error into `(nil, nil)`, so a DB failure is indistinguishable from a deleted provider and silently disables OCR/AI.
5. Changing the OCR provider in Settings doesn't clear the model (the OCR test page does), so e.g. `mistral-ocr-latest` can be saved against an OpenAI provider and only fails at processing time.
6. `ProviderModelFields` renders no placeholder option when the binding is empty but providers exist, so the UI shows a provider that isn't actually in form state.

### Non-blocking notes

- `resolveProviders` writes the chat/search fallback back into `Config`, which `settingsResponseFromConfig` returns and the Settings form saves verbatim. The "chat and search inherit extraction" behavior therefore becomes a hard binding after the first save. Intentional? If so it's worth saying in the UI; today the field just silently fills in.
- `EnsureUniqueAlias` swallows the lookup error (see inline), so duplicates fall through to the unique index and surface as a raw `Failed to create provider: ...` 400.
- `handleCreateProvider` / `handlePatchProvider` concatenate `err.Error()` from `app.Save` into the response. That's internal validation/SQL detail; admin-only, so low risk, but inconsistent with the other handlers here.
- `llmOCRMaxFileBytes` is 10 MiB of raw bytes, which becomes ~13.3 MiB on the wire after base64. Worth confirming that's under the limits of the providers you expect people to use.
- The PR description leaves the manual test checklist and the Docker build unchecked. The Docker build in particular matters since `AGENTS.md` requires it for build-affecting changes.

backend/internal/appapi/providers.go
```diff
@@ -0,0 +1,188 @@
...
+	return func(e *core.RequestEvent) error {
+		id := strings.TrimSpace(e.Request.PathValue("id"))
+		p, err := aiprovider.FindByID(app, id)
+		if err != nil || p == nil {
+			return writeError(e, http.StatusNotFound, "Provider not found.")
+		}
+		forOCR := strings.EqualFold(strings.TrimSpace(e.Request.URL.Query().Get("for")), "ocr")
+		models, err := aiprovider.ListModels(e.Request.Context(), *p, forOCR, nil)
+		if err != nil {
+			return writeError(e, http.StatusBadGateway, err.Error())
```
buldezir: `err.Error()` here carries the upstream response body. `ListModels` builds it as:

```go
return nil, fmt.Errorf("list models: HTTP %d: %s", resp.StatusCode, msg)
```

where `msg` is up to 300 bytes of the provider's raw error JSON. OpenAI and OpenRouter both include account-scoped detail in those payloads (organization IDs, key-permission and quota messages). Combined with this route being `bindAuth`, that reaches any logged-in user.

Returning a generic message and logging the detail server-side would be enough — the frontend already falls back to "You can still type a model id" and doesn't need the upstream text to be useful:

```go
if err != nil {
	app.Logger().Warn("list provider models", "provider", p.ID, slog.Any("error", err))
	return writeError(e, http.StatusBadGateway, "Failed to load models from the provider.")
}
```

backend/internal/appapi/providers.go
```diff
@@ -0,0 +1,188 @@
...
+		if err := json.NewDecoder(e.Request.Body).Decode(&req); err != nil {
+			return writeError(e, http.StatusBadRequest, "Invalid request body.")
+		}
+		sdk := record.GetString("sdk")
+		if req.SDK != nil {
+			sdk = strings.TrimSpace(*req.SDK)
+			if !aiprovider.ValidSDK(sdk) {
+				return writeError(e, http.StatusBadRequest, "sdk must be openai, openrouter, google_vision, or mistral.")
+			}
+			record.Set("sdk", sdk)
```
buldezir, line 135: Changing `sdk` isn't checked against the bindings that already point at this provider, which makes it possible to break the install from the Settings UI with no warning.

Say `extract_provider_id` points at this record and it's currently `openai`. Editing it to `mistral` saves fine. Then `Runtime.apply` skips building the extractor because `aiprovider.IsLLM` is false, `HasLLM(cfg)` returns false, `needsConfigSetup` returns true — and the SPA hard-gates every user back into the setup wizard until an admin figures out what happened.

`applySettingsPatch` already enforces exactly this rule from the other direction via `validateProviderID(app, id, true)`; the same invariant just isn't enforced when the provider mutates underneath a stable binding. Worth loading the settings record and rejecting the change:

```go
if !aiprovider.IsLLM(sdk) {
	if settings, err := config.FindSettingsRecord(app); err == nil {
		for _, field := range []string{"extract_provider_id", "chat_provider_id", "search_provider_id"} {
			if strings.TrimSpace(settings.GetString(field)) == record.Id {
				return writeError(e, http.StatusConflict, "Provider is bound to extraction, chat, or search and must stay openai or openrouter.")
			}
		}
	}
}
```

That mirrors the 409 that `handleDeleteProvider` already returns for in-use providers, which is the behavior an admin would expect here too.

backend/internal/config/config.go
```diff
@@ -254,22 +285,84 @@ func configFromRecord(record *core.Record) Config {
...
+}
+
+func lookupProvider(app core.App, id string) (*aiprovider.Provider, error) {
+	p, err := aiprovider.FindByID(app, id)
+	if err != nil {
+		if strings.TrimSpace(id) == "" {
+			return nil, nil
+		}
+		// Missing record is a soft miss so the process stays up.
+		return nil, nil
```
buldezir: Both branches return `(nil, nil)`, so the `id == ""` check above is dead — `FindByID` already returns `(nil, nil)` for an empty id without touching the DB, meaning `err` is never non-nil in that case.

More importantly, collapsing every error into a soft miss means a genuine DB failure looks identical to a deleted provider: `Load` succeeds, `Runtime.apply` builds no clients, and the app quietly drops into "not configured" with only the `runtime settings reloaded` line (`"ocr": "unavailable"`) as evidence. The comment says the intent is tolerating a *missing record*, so it'd be worth narrowing to that and logging the rest:

```go
func lookupProvider(app core.App, id string) (*aiprovider.Provider, error) {
	p, err := aiprovider.FindByID(app, id)
	if err != nil {
		if errors.Is(err, sql.ErrNoRows) {
			return nil, nil
		}
		return nil, fmt.Errorf("lookup provider %s: %w", id, err)
	}
	return p, nil
}
```

If you'd rather keep the process up unconditionally, at least log at warn level so a broken binding is diagnosable from the logs.

backend/internal/aiprovider/provider.go
```diff
@@ -0,0 +1,93 @@
...
+
+func EnsureUniqueAlias(app core.App, alias, excludeID string) error {
+	alias = strings.TrimSpace(alias)
+	if alias == "" {
+		return fmt.Errorf("alias is required")
+	}
+	record, err := FindByAlias(app, alias)
+	if err != nil {
+		// FindFirstRecordByFilter returns an error when no record matches.
+		return nil
```
buldezir, line 75: The comment explains why the error is ignored, but ignoring it also swallows real failures: any DB error makes this report "alias is free" and the caller proceeds to `app.Save`, where the unique index rejects it and the admin gets `Failed to create provider: <raw validation error>` instead of the clean `alias %q is already in use` message this function exists to produce.

Since `FindFirstRecordByFilter` signals "not found" specifically with `sql.ErrNoRows`, you can distinguish the two:

```go
record, err := FindByAlias(app, alias)
if errors.Is(err, sql.ErrNoRows) {
	return nil
}
if err != nil {
	return fmt.Errorf("check alias: %w", err)
}
```

Minor, but it's the difference between a comprehensible 400 and a confusing one.

frontend/src/components/ProviderModelFields.tsx
```diff
@@ -0,0 +1,138 @@
...
+  return (
+    <div className="grid gap-4 sm:grid-cols-2 sm:col-span-2">
+      <label className={labelClassName}>
+        <span className={labelTextClassName}>{label} provider</span>
+        <select
+          className={inputClassName}
+          value={providerId}
+          onChange={(event) => onProviderChange(event.target.value)}
+        >
+          {allowEmpty || providers.length === 0 ? <option value="">Select a provider</option> : null}
```
buldezir: When `allowEmpty` is false and `providers.length > 0` but `providerId` is `''`, the `<select>` has no option matching its value, so the browser displays the first provider while React state stays empty. Saving then submits `ocr_provider_id: ''` / `extract_provider_id: ''` — which `validateProviderID` accepts as "unset" — and the admin ends up with nothing bound despite the form showing a selection.

This is reachable on an install with no env-seeded keys: bindings start empty, the admin adds a provider in the Providers section above, and the Models section immediately renders in this state.

Rendering the placeholder whenever there's no selection fixes the display, and the caller-side `if (!ocrProviderId) throw` guards in `SetupWizard` already assume that shape:

```tsx
{allowEmpty || providers.length === 0 || !providerId ? (
  <option value="">Select a provider</option>
) : null}
```

frontend/src/routes/settings.tsx
```diff
@@ -229,127 +285,191 @@ export function SettingsPage() {
...
-                onChange={(e) => updateField('openai_model', e.target.value)}
-              />
-            </label>
+            <ProviderModelFields
+              label="OCR"
+              providers={providers}
+              providerId={form.ocr_provider_id}
+              model={form.ocr_model}
+              purpose="ocr"
+              onProviderChange={(id) => updateField('ocr_provider_id', id)}
```
buldezir, line 429: Switching the OCR provider leaves `ocr_model` untouched, so a model id from the previous provider carries over. Going from a Mistral provider to an OpenAI one keeps `mistral-ocr-latest`, and nothing rejects it: `applySettingsPatch` only checks that `ocr_model` is non-empty, and `NewFromAIProvider` only checks the same. The first document to hit the pipeline fails with an upstream "model not found", far from where the mistake was made.

The OCR test page already handles this — `ocr-test.tsx` does `setProvider(...)` and `setModel('')` together in its `onChange`. Same treatment here:

```tsx
onProviderChange={(id) => {
  updateField('ocr_provider_id', id)
  updateField('ocr_model', '')
}}
```

The extraction/chat/search fields below have the same issue, though it's less likely to bite since they're all filtered to LLM providers with overlapping catalogs. Worth doing consistently.

backend/internal/worker/processor.go
buldezir: Fixed in 3759b58.

backend/internal/appapi/appapi.go
buldezir: Fixed in 3759b58.

backend/internal/appapi/providers.go
buldezir: Fixed in 3759b58.

backend/internal/appapi/providers.go
buldezir: Fixed in 3759b58.

backend/internal/config/config.go
buldezir: Fixed in 3759b58.

frontend/src/components/ProviderModelFields.tsx
buldezir: Fixed in 3759b58.

frontend/src/routes/settings.tsx
buldezir: Fixed in 3759b58.

Issue #4, opened 2026-08-08
docker hub or github for prebuilt docker image


cirrusflyer:
Consider adding images to Docker Hub or Github?

buldezir:
@cirrusflyer lol mate, in "packages" from day zero
https://github.com/buldezir/paperless-go/pkgs/container/paperless-go

later i will add them to compose and docs

cirrusflyer:
My bad - just was noticing the compose file.  Thanks!

PR #35, opened 2026-09-01
Rework the AI environment into a bootstrap, and give hosting an AI_MANAGED mode

buldezir:
Paired with **lemmary/lemmary-dev#4**, which carries the e2e updates. That side is already pushed, so the pull request job will pick up the matching suites.

## Why

AI was configured three times over. `.env` carried nineteen AI-shaped variables across three SDK-specific families, a first-run wizard collected the same thing, and the Settings page collected it again — and every one of those variables already had a database column behind it. Nothing in the runtime path reads the environment: `internal/ai`, `internal/ocr` and `internal/worker` take everything from `config.Snapshot`. The environment layer was pure seeding.

And it had grown subtle enough that a second repository had to mirror it: `lemmary-panel`'s `TenantSeedGaps` reimplements `aiprovider.SeedFromEnv` plus `HasLLM`/`HasOCR`, with a comment admitting the mirror can drift because the original lives elsewhere.

Two things should be true instead. Self-hosted: the environment is a convenience for the *first* boot — enough to bring an instance up ready to use without walking the wizard — and Settings owns it after that. Hosted: the operator carries the AI bill, so the environment is mandatory and authoritative and the tenant's Settings page does not offer AI at all. One build, one image; the difference is a runtime flag, per AGENTS.md *"One build, feature flags in the environment"*.

## The surface

```sh
AI_SDK / AI_API_KEY / AI_MODEL / AI_BASE_URL       # openai | openrouter | mistral
OCR_SDK / OCR_API_KEY / OCR_MODEL / OCR_BASE_URL   # + google_vision; unset = OCR runs on the AI provider
AI_MANAGED
SETUP_ADMIN_EMAIL / SETUP_ADMIN_PASSWORD
```

One key is usually the whole configuration: `AI_API_KEY` alone creates one provider and routes extraction, chat, Deep Search **and** OCR to it. Naming the same SDK for both reuses that key, endpoint and provider *record* — the same endpoint under a second alias would be a duplicate row the Settings page could not explain.

**Retired outright** (clean break, no shims): `OPENAI_API_KEY`, `OPENAI_BASE_URL`, `OPENAI_MODEL`, `OPENAI_CHAT_MODEL`, `OPENAI_SEARCH_MODEL`, `MISTRAL_API_KEY`, `MISTRAL_API_BASE_URL`, `MISTRAL_MODEL`, `MISTRAL_OCR_MODEL`, `GOOGLE_VISION_API_KEY`, `OCR_PROVIDER`, `PROCESSING_RESULT_LANGUAGE`, `VITE_DEV_USER_EMAIL`, `VITE_DEV_USER_PASSWORD`. `OPENAI_TIMEOUT_SEC` → `AI_TIMEOUT_SEC`.

The three separate model pins go with them — one model serves extraction, chat and Deep Search, so the fallback chain every reader had to hold in their head has nothing left to resolve. The per-task bindings still exist in the database and Settings still sets them independently; only the *environment* stopped speaking in four models.

`PROCESSING_RESULT_LANGUAGE` was dropped rather than renamed: it decides what language a document's title and summary are stored in, which is a reader's preference, not an operator's.

## The two modes, and what replaced `env_applied`

| | self-hosted (default) | managed (`AI_MANAGED=1`) |
|---|---|---|
| env → DB | first boot only | **every** boot |
| authority after | Settings page | the container's environment |
| Providers/Models/Duplicates | editable | not rendered, endpoints `403` |
| incomplete or invalid | wizard opens | **refuses to start**, naming the variable |

`app_settings.env_applied` held a SHA-256 digest per variable so a *changed* variable was applied and an unchanged one left alone. That existed only to let one code path serve both an orchestrator and an admin. Naming the modes makes it unnecessary — neither asks whether anything changed — so `envapply.go` and its test are deleted and migration `1730000018` drops the column.

Refusing to start is right in managed mode and only there: nobody inside a managed instance can repair a bad key, and the panel's create ladder ends in a health check, so a bad seed fails the provision instead of handing over a broken workspace. It is parsed in `main.go` beside `boot.Prepare`, before `pocketbase.New` — a pure read with no database behind it.

This also closes a gap the old design recorded as deliberately unsolved: nothing stopped a hosted user editing the model in Settings to escape their plan's routing. That lock is now in core behind a runtime flag rather than owed to a private edition seam.

## What managed takes, and what it leaves

`ApplyManaged` writes the providers, the four bindings and their models, `SEARCH_CONTEXT_TOKENS` and `NEAR_DUPLICATE_*` — each a cost the operator carries, and near-duplicate is priced per tier by the panel's plan ladder. Timeouts, retries, languages and the prompt version are deliberately **not** in that set: tuning with no operator decision behind it, and resetting somebody's timeout on every restart of a container they do not control would be a bug rather than a policy.

Hiding the sections in the SPA is a courtesy, not the guard. The endpoints are reachable with any admin session regardless of what the page rendered, and a tenant who swapped in their own key would move the AI bill onto an account the operator is billing them for — so the writes refuse server-side.

## Settings: Models is split, not hidden whole

Hiding Models wholesale would also have hidden the result language, the timeouts and the deep-search languages, which a managed tenant keeps. The page is now **Providers / Models / Processing / Worker / Duplicates**, where a hidden section means exactly *"the environment owns this here"*. The patch body matches, omitting managed fields rather than sending them unchanged — the server refuses the whole patch if it names one, so including them would fail the save of the fields still editable.

## `SETUP_ADMIN_*`, and why `VITE_DEV_USER_*` folded into it

With these beside an AI key, `docker compose up` on a fresh volume comes up with nothing left to answer. Create-if-absent, never an upsert: upserting would reset the password on every restart of a long-lived container, silently undoing an admin who had changed it. The wizard's superuser + paired-`users` creation is factored into `appapi.CreateFirstAdmin` so the two callers cannot drift — the pairing matters, since a superuser without a `users` record cannot own a document they upload.

`VITE_DEV_USER_*` was the same credential pair under a second name. The dev auto-login now reads `SETUP_ADMIN_*` directly via `envPrefix` in `vite.config.ts`, because `CreateFirstAdmin` makes the very `users` account `devCredentials()` signs into. The `import.meta.env.DEV` guard is unchanged and still keeps the pair out of a production bundle.

## Beyond the plan: `OCR_BASE_URL`

`AI_BASE_URL` existed with no OCR counterpart — an asymmetry rather than a decision. The e2e harness proved it: it points OCR at a different mock server than the language model and had no way to say so.

## Verification

`./scripts/test-all.sh` passes in full against lemmary/lemmary-dev#4 — Go units, vet, frontend tests, the SPA build, API e2e and 71 browser specs — and `docker build` succeeds.

Checked against the real image on a shared volume:

```
boot1 (self-hosted, seeded):        first-model
after a Settings edit:              chosen-in-settings
boot2 (self-hosted, env changed):   chosen-in-settings   ← the new rule
boot3 (managed):                    operator-model       ← reclaimed
```

`AI_MANAGED=1` with no key exits 1 before serving a request; with one, `PATCH extract_model` and `POST /providers` answer `403` while `worker_max_retries` still saves. A fresh volume with `SETUP_ADMIN_*` reports `needs_admin:false, needs_config:false` and both accounts sign in. A production bundle contains neither credential.

**Not covered:** no browser spec for the managed Settings page rendering — the Playwright suite runs one server, self-hosted. The hiding is typechecked and the server-side refusal is e2e-tested.

## Follow-up, not in this PR

`lemmary-panel` still names the retired variables in `tenant.AllowedEnvKeys`, `config.TenantSeedGaps`, `compose.dev.yml` and the plan ladder's per-tier model columns. Its `Env` is inside `SpecHash`, so that change recreates the fleet on the next reconcile and wants doing deliberately.

Anyone with an existing `.env` needs to migrate it by hand — a clean break means a stale file silently lands on the setup wizard (or, under `AI_MANAGED=1`, refuses to boot, which is the loud case).


buldezir:
Worked through all eight. **Six were real and are fixed** in fb4f686; two I'm pushing back on, with reasons.

I reproduced each before changing anything rather than taking them on trust — the repros are in the commit message.

## Fixed

**1 — named OCR provider with no key.** Confirmed exactly as described: `AI_MANAGED=1 AI_API_KEY=sk OCR_SDK=google_vision` parsed clean, `SharesOneProvider()==true`, `OCRSDK()=="openai"`. The root cause is that `Configured()` means "has a key" and `SharesOneProvider()` was built on it, so an SDK named without a credential was indistinguishable from no OCR provider at all — which is also why my own managed check could never fire.

Two changes: `ProviderSpec.Requested()` (was an SDK named) now drives the OCR resolution, and `parseOCR` refuses a second endpoint with no key **in both modes**. You noted the self-hosted mis-bind too, and I think it's the worse half — the env seeds once and goes inert, so it's permanent and the wizard never asks. The managed clause is deleted rather than fixed: with the parser refusing that input it's genuinely unreachable, and an unreachable check reads as protection that isn't there.

**2 — `ApplyManaged` warned and carried on.** Agreed, and your framing is the right one: "the app must come up so an admin can go and look" is correct everywhere else in that hook and wrong here, because in managed mode there's nobody to look. It fails the boot now.

**3 — `AI_MANAGED` fail-open.** Confirmed: `yes` and `on` both ran as self-hosted — and those are spellings `.env.example` documents as valid for `VAULT_*` a few blocks further down the same file, which makes it worse than a typo hazard. Now uses `vault.envBool`'s treatment. I left every other fallback alone; a typo in a timeout must not become a zero-second deadline, but that argument doesn't transfer to a lock.

**4 — SPA fails open until meta resolves.** `aiManaged` is tri-state now (`undefined` = not known or not answerable), and Settings treats anything but an explicit `false` as managed. Worth being precise that this was never a hole — the server refuses regardless — but your second-order point is the real bug: because `touchesManaged` 403s the *whole* patch, a meta blip made the timeouts and languages a managed tenant genuinely owns unsavable.

**5 — shipped superuser password.** The one I'd have most regretted. Those values carried over from `VITE_DEV_USER_*`, which could never do more than fill in a form in a dev build; they now mint a real `_superusers` account. Anyone following the documented `cp .env.example .env` got an admin whose password is published in this repo. Commented out, with the reason where someone is deciding whether to uncomment.

**6 — sentinel in the wizard message.** Confirmed: `invalid admin credentials: Email is required.` A small `invalidAdmin` type carries the sentence and implements `Is()`, so `errors.Is` still separates bad input from a database failure.

## Not changing

**7 — read-side managed lock.** I checked both: `duplicates.ScanAll` (`internal/duplicates/scan.go:118`) is DB reads, checksums and local text comparison with no AI call anywhere in the package — its cost is CPU, already bounded by the container's plan limits. `GET /providers/{id}/models` is a catalog listing, free on these providers, and only reachable from a UI that managed mode doesn't render. Neither is billed AI work, so gating them would add surface without closing anything. Happy to revisit if you're seeing a provider that charges for model listing.

**8 — missing tests.** Half right, and the half that's wrong isn't your fault: `ApplyManaged` (reclaim, idempotence across three applies, key rotation), both `403`s, the `ai_managed` flag and `CreateFirstAdmin`'s create-if-absent are all covered — in the private `lemmary-dev` suites, which a PR diff can't show. See lemmary/lemmary-dev#4.

Genuinely missing, and now added: the two parser inputs that would have caught findings 1 and 3, every documented boolean spelling both ways, and `aiprovider/bootstrap_test.go` for the pure resolution across all four shapes — including the hand-constructed SDK-without-key that `parseOCR` now refuses, since nothing in that package enforces the parser's guarantee.

You were right that finding 1 was a one-liner in `aienv_test.go`. What actually caught the fallout was the e2e harness: `EmptyAPIKeys` was itself producing the half-written input, so every setup-wizard suite failed to start the moment the app began refusing it. The parser tests state their whole input and couldn't have.

## Verification

`./scripts/test-all.sh` green in full, image rebuilds. Against the real image:

```
AI_MANAGED=1 + OCR_SDK=google_vision, no key  -> exits, names OCR_API_KEY
AI_MANAGED=yes                                -> honoured; goes on to require AI_API_KEY
AI_MANAGED=ture                               -> refuses to start
OCR_SDK=google_vision + key                   -> OCR bound to Vision, extraction to OpenAI, no model on Vision
```

PR #36, opened 2026-09-01
Add a timeline sidebar to the documents page

buldezir:
Adds a timeline sidebar to the documents page: a per-month histogram of the owner's documents that doubles as a date filter.

## Backend
- `GET /api/app/documents/timeline` returns the owner's document counts grouped by month of `document_date` (`appapi/documents_timeline.go`).
- Migration `1730000017_document_date_index` adds a covering `(user, document_date)` index. Without it the GROUP BY scans the base table, which stores `ocr_text` inline — up to `MaxOCRTextRunes` per row. The index also covers the From/To filters on the documents list, which previously fell back to `idx_documents_user_created` and filtered afterwards.

## Frontend
- `DocumentTimeline.tsx` renders the histogram; clicking a month sets the list's date range.
- `lib/timeline.ts` holds the bucketing and range math, unit-tested in `lib/timeline.test.ts`.
- `routes/index.tsx` grows the sidebar layout.

## Notes
- The migration shares the `1730000017` prefix with `search_context_tokens` from main. Harmless: PocketBase tracks migrations per filename, the two touch different collections, and the repo already has a duplicate prefix at `1730000014`.
- E2E and browser coverage is the companion PR in `lemmary-dev` on the same branch name.

## Verification
- `tsc -b` clean; 105 frontend tests pass
- `go build ./...` and `go test ./...` clean


buldezir:
Good catch on the blocking one — confirmed and fixed in 715ece5.

`document_date` is a `core.DateField` (`migrations/1730000001_initial.go:96`), so it stores `YYYY-MM-DD HH:MM:SS.sssZ` and compares as a string, and `buildDocumentFilter` asked for `document_date <= '2025-03-31'`. Meanwhile `fulltext.parseDayBoundary` already rolls `date_to` to an exclusive next-day bound, so the two paths genuinely disagreed on the same From/To pair. The list now bounds it the same way (`document_date < dayAfter(dateTo)`); `monthRange`/`yearRange` still return the inclusive last day, since that is what the date input shows.

Tests: `buildDocumentFilter` now asserts the emitted bound against a DateField-shaped `'2025-03-31 00:00:00.000Z'` — the string compare that was failing — plus month-end, year-end, and leap-day rollover, and that an unparseable To passes through rather than inventing a bound.

Sidebar clipping: fixed too, `lg:max-h-[calc(100vh-4rem)]` + `lg:overflow-y-auto`.

On backend coverage of the SQL and the owner `WHERE`: that exists, just not in this repo. The e2e suites live in a separate private repository and the companion PR there covers the endpoint end to end, including month bucketing, owner scoping, and that undated documents stay out of the histogram. `buildTimeline` is unit-tested here because it is the part that is worth isolating from a live database.

PR #38, opened 2026-09-01
Deep Search: hybrid passage retrieval, embeddings, and topic-scale Research with a helper model

buldezir:
## Summary

Deep Search moves from document-level keyword lookup to passage-level hybrid retrieval, and Research from reading documents one at a time to covering whole topics at a cost that grows with the documents touched rather than with the square of them. BM25 and dense vectors are fused with reciprocal rank fusion; long documents are read as excerpts; bulk per-document work goes to a cheaper helper model; the agent gains survey and count tools; and evidence carries across conversation turns. The Documents page search is unchanged.

## Retrieval

- **Hybrid `search_documents`.** Relaxed BM25 over the documents index (min-should-match rungs, one edit of fuzziness on longer words, no stemmers by design) fused with kNN over embedded passages by reciprocal rank fusion. Every hit carries one to three verbatim passages chosen from the chunk index, or cut from the OCR text when there is none. Filters (type, correspondent, tag, date) are resolved against the documents index and passed to the passage search as ids; a filter name that resolves to nothing matches nothing rather than everything.
- **Reads are excerpts.** `read_documents` returns a long document as its head plus the passages ranked against `focus`, or against the user's question when the model named none, with gaps marked. The whole text of a long document is never handed to the research model.
- **Prior evidence.** A follow-up question can read documents cited earlier in the same chat by id without searching again; they join the answer's document list only when cited.
- **Language prompt follows retrieval.** With an embedding model bound the prompt says one search already crosses languages and forbids translated repeats; the `DEEP_SEARCH_LANGUAGES` list only drives per-language searches in keyword-only mode. The Settings hint changes accordingly.

## Research at topic scale

- **Deep Search helper.** A sixth AI binding (`search_helper_provider_id` / `search_helper_model`, seeded by `AI_SEARCH_HELPER_MODEL`, operator-owned under `AI_MANAGED=1`) for many cheap per-document calls. Unset, it falls back to the Search model. JSON mode a provider rejects is retried in plain text and parsed leniently.
- **Distilled reads.** A read past 32 KB of text or five documents is read by the helper, which returns per-document notes, verbatim quotes and requested values; smaller reads pass through as text. The helper sees each document whole up to 400 KB and several short documents share one call. A helper failure falls back to excerpts.
- **`survey_documents`.** The same retrieval as a search, kept to 300 documents by default and 1000 at most, every document read by the helper for one question, one compact row back per document. Number fields are summed, averaged and bounded on the server per currency; the model is told to report those figures rather than add rows itself. Progress streams as "Surveyed 120 of 300 documents".
- **`count_documents`.** How-many and distribution questions answered without reading: filters alone by SQLite (`COUNT(*)`, `GROUP BY` type, correspondent, year, month or tag; undated documents excluded from ranges; legacy tag values guarded), query text by the index's exact strict-match total, grouped text counts by taking the index's ids to the database (approximate past 5000 matches). Unresolved filter names come back by name.
- **Token usage in logs.** Every completion logs prompt, cached and completion tokens (streaming requests ask for the usage chunk), and a research run logs its total.
- Research has no round or document cap; the loop ends when the model is ready, stalls, or the provider rejects the request. DSML-style models still get one tool round.

## Embeddings

- Fifth AI binding: embedding provider and model, seeded by `AI_EMBEDDING_MODEL`, operator-owned under `AI_MANAGED=1`. Unset means dense retrieval is off and behaviour is the keyword search every install had. The recorded vector length is read-only in Settings.
- Documents are chunked deterministically (byte offsets, versioned) plus a metadata header passage, embedded through the OpenAI-compatible `/embeddings` endpoint in batches, and stored as float32 vectors in raw SQLite tables inside the database (so inside the vault). A soft-failing `embed` pipeline step and a backfill cron cover new uploads, existing archives, restores, model switches and retries; renaming a tag, correspondent or type marks referencing documents stale because the header embeds names.
- Management gains an Embeddings section with stored/pending/failed counts and an "Embed N missing documents" sweep (`POST /api/app/embeddings/backfill`, admin, shares the cron's lock). Any document not currently processing can be reprocessed from its page, including restored documents that never had a job.
- OpenRouter's embedding picker asks the catalogue with `output_modalities=embeddings`; the plain listing omits embedding models entirely.

## Bleve native vector search

- A second derived index, `bleve/chunks`, holds passage text and vectors (cosine, `vector_base64`), built per model and dimension and rebuilt from SQLite on boot or unlock with no embedding calls. kNN with owner and document pre-filtering; chunk-level text plus kNN uses Bleve's native RRF. Orphan sweeps and chunk replacements notify the index.
- This needs the `vectors` build tag and cgo against the blevesearch FAISS fork. The image moves from Alpine to Debian trixie with a cross-compiled FAISS stage; `-tags vectors` is always on and a `!vectors` sentinel makes a tag-less build fail with one readable error. `scripts/faiss-build.sh` is the single source of the FAISS pin; `docs/setup.md` covers the local toolchain.

## Cost and operations

- Image about 386 MB (was about 130 MB); the first FAISS build takes a few minutes per architecture, then is cached.
- Embedding cost is documented in `.env.example` and `docs/setup.md` (tokens per document, re-embed triggers, about 6 KB per chunk at 1536 dims). Under `VAULT_ENABLED=1` vectors and the chunk index live in tmpfs; prefer models of 1024 dims or fewer.
- Helper calls run four at a time; survey rows are on the order of 100 tokens each, so a full 1000-document survey is roughly 100k tokens back to the research model once.

## Verification

- `./scripts/test-all.sh`: Go tests and vet with `-tags vectors`, frontend tests and build, overlay API e2e and browser e2e. New tests cover the helper's JSON parsing and JSON-mode fallback, distillation thresholds and fallback, survey totals and number parsing, the count SQL against in-memory SQLite, tool dispatch in the research loop, and the dense and keyword-only prompt variants.
- `docker build -t lemmary:local .`; `ldd` on the binary resolves `libfaiss_c`; the container adopts a root-owned volume and reports healthy. arm64 cross-compiles; the full arm64 image needs QEMU binfmt (CI has it).
- Eval floors recorded in `fulltext/eval_test.go` and `retrieval/eval_test.go`.
- Not exercised in this branch: a live end-to-end run of survey and count against a real provider. Compare the logged `research run usage` totals for a broad question before and after.


buldezir:
Addressed all review findings in the latest commit:

1. Focused/sequential reads now use the raw `ocr_text` column, the same coordinate system `chunk.Split` used (test with leading whitespace added).
2. Whitespace-only OCR: SQL predicate trims, and `EmbedDocument` writes a terminal row when there is nothing to embed, so a document cannot stay a candidate across ticks.
3. Renaming a tag, correspondent or document type marks referencing documents stale; `staleFields` also gains `tags` (which the header embeds and was missing) plus `purpose_original`/`summary_original`, now included in the header passage.
4. Stored passage text cap derived from the chunker and header limits (2000 runes); `ChunkMappingVersion` bumped so existing indexes rebuild from SQLite.
5. Provider SDK switch refused when bound only as the embedding provider.
6. `DeleteOrphans` returns the swept ids and notifies the chunk index.

`./scripts/test-all.sh` passes end to end (Go + vet with `-tags vectors`, overlay API e2e, frontend, 86 browser e2e).

buldezir:
Merged `main` into this branch. Per the resolution: the search functions keep this branch's design (passages, hybrid retriever, prior-turn evidence), the context-window budget is removed as on main (AI requests are bounded only by the provider; `read_documents` keeps `focus`, drops `offset`/`next_offset`), and CI/verification follow main's Docker-based flow. The overlay branch `lemmary-dev@deep-search-rag` adds FAISS and `-tags vectors` to the verify image; both the host and the Docker-based `./scripts/test-all.sh` pass end to end.

buldezir:
Follow-up: documents without a processing job (restored archives, pre-pipeline documents) now show the processing panel and can be reprocessed, so the `embed` step is reachable for them. Management gains an **Embeddings** section with stats and an "Embed N missing documents" action (`POST /api/app/embeddings/backfill`, admin; `GET` polls progress). Overlay branch updated with API and browser e2e (the mock provider now serves `/embeddings`). Full Docker-based suite green (90 browser specs).

PR #43, opened 2026-09-04
Run the embedding model yourself, in a sidecar

buldezir:
Turning `AI_EMBEDDING_MODEL` on has meant sending every passage of every document to a hosted provider, forever, and re-sending the whole archive whenever a title is edited or the model changes. That is a recurring bill and a privacy decision, and embeddings are the one AI job here that does not need a frontier model: a 500 MB – 2 GB model on the host's own CPU is competitive with `text-embedding-3-small`.

An operator can now run [text-embeddings-inference](https://github.com/huggingface/text-embeddings-inference) as a compose sidecar and point Deep Search's dense half at it. Nothing else moves — OCR, extraction, chat and the Search agent stay on whatever `AI_SDK` names — and absent, everything behaves exactly as before.

## A new keyless SDK, `local`

An OpenAI-compatible `/v1/embeddings` endpoint you run yourself. It embeds and nothing else — refused as `AI_SDK` and `OCR_SDK` — and it needs no credential, because a service on the compose network has nobody to authenticate to.

Adding it forced apart three questions `IsLLM` had been answering at once, because until now every SDK that chatted also embedded. `aiprovider` now has `CanEmbed`, `CanOCR` and `RequiresAPIKey`, and each binding asks the one that applies to it. That also catches two bugs this change would otherwise have introduced: `ValidSDK` is not the question OCR asks, so `OCR_SDK=local` and an OCR provider binding both passed validation and would have failed on the first uploaded document. Both are now refused at configuration time.

## Merging #42

`docling` (keyless OCR sidecar) and `local` (keyless embeddings sidecar) arrived from opposite ends and grew their own answer to the question `"has an API key"` used to stand in for. Most of the merge was reconciling duplicates rather than resolving text — `HasCredential` dropped for main's stricter `Provider.Configured`, `ProviderSpec.Usable` with it, `sdkRequiresKey` for `requiresAPIKey`, and a hardcoded `invalidSDKMessage` for main's derived one.

Three things needed more than picking a side:

- `ListModels` guarded on `!IsLLM`, which silently short-circuited the local sidecar before it could read TEI's `/info`. It now guards on "neither chats nor embeds".
- `handleOCRProviders` had stopped filtering on the API key so docling would show; unfiltered it would also have listed a `local` provider on the OCR test page.
- **Both branches added a migration widening `ai_providers.sdk` at the same timestamp** (`1730000024`). Collapsed into one: main's filename, this branch's idempotent `ValidSDKs`-derived body, and a down-migration that clears the embedding binding as well as the OCR one.

## Sized against measured memory, not a misread flag

The sidecar crash-looped on a 16 GB host: OOM-killed during warmup at 11.6 GB. `--max-batch-tokens=65536` was set on the mistaken reading that it guards the `413` a too-small server returns on a full batch. It does not — that is `--max-client-batch-size` plus `--auto-truncate`. It is TEI's forward-pass budget, and on CPU it is what decides resident memory.

| `--max-batch-tokens` | result | peak RSS |
| --- | --- | --- |
| 65536 / 16384 / 8192 | OOM-killed | 11.6 GB (host ceiling) |
| **6144** | **ready** | **~8.1 GB** |
| 4096 | ready | ~6.5 GB |

6144 rather than 4096, because the floor is what one input can be: `embed.go` caps a single input at 8000 runes, which is 5723 tokens in the densest script measured. Below that TEI truncates in silence — at 4096 a 5723-token input embeds *byte-identically* to its first 4096 tokens, `200`, no warning. `mem_limit: 10g` bounds the failure, which was a *global* OOM: with no limit the kernel picks the largest process, and on a busier host that could have been the app or the database.

## Two UI bugs, both older than the timing work

- **Settings filtered `local` out of its own embedding picker.** It built `providers.filter(isLLMProvider)` and passed it to five of six pickers; `local` embeds without chatting, so the one binding it exists for was the one it could not reach. Filtering now lives in `eligibleProviders`, and no call site pre-narrows its list.
- **The document page stopped polling on the document's status.** `apply_metadata` marks the document `completed` and saves it, and *then* embed runs — 8.2 s in the job I traced. The panel froze with embed reading `running` for ever. The gate now also asks the job, keyed on `finished_at`, the only field that means the whole job is done.

## Also

Each processing step now shows how long it took, and the job a total — two timestamps that were already on the wire, never rendered. Skipped steps report no duration: the pipeline decides to skip before booking an attempt, so they carry a finish and no start, and the obvious subtraction would print the Unix epoch.

`docs/local_embeddings.md` gives the feature its own page beside `local_ocr.md`.

## Verified

Backend `build`/`vet`/`test -tags vectors` clean; 174 frontend tests. The sidecar was run for real: ready in ~25 s, healthy, 7.9 GB of its 10 GB limit, serving 1024-dim vectors, and not truncating a 5723-token chunk.

## Left for a decision

- `docker-compose.embeddings.yml` publishes `8999:80`, contradicting the "No `ports`" comment directly beneath it — an unauthenticated inference endpoint on the host. Left in place at the author's request; the docs now say so plainly rather than claiming otherwise.
- That overlay's `app:` block is commented out, so unlike the GPU one it does **not** bind the app. Documented as-is rather than changed.
- A document is marked `completed` while a step is still running. Defensible — its text and metadata are final, embeddings only enrich retrieval — but it is why a document's status and its job's status are two different questions.


buldezir:
All four findings were real and are fixed in 0c95cd6. Verified each against the code or the registry first.

**1 — concurrent reprocess during embed (high).** Confirmed, and it had a third face the review did not mention. `apply_metadata` sets the job's *status* to completed and saves it before embed runs, so for that window a live job is `status=completed` with `finished_at` empty. All three call sites now ask `finished_at`, which is written exactly once on both terminal paths and deliberately left empty by a retry:

- `createProcessingJob`'s guard
- `canReprocess` on the document page
- `recoverStaleRunningJobs` — the same blind spot, and the two interact badly: it re-pends `status=running`, so a crash *during* embed left a job it would never recover, and the tightened guard would then have blocked that document from reprocessing for ever.

The regression test is DB-backed rather than a filter-string assertion, since the bug lives in SQL. It fails against the old guard and passes against the new one, with the states either side of the window covered so the guard cannot just refuse everything.

**2 — unauthenticated TEI (high).** Confirmed. Bound to `127.0.0.1` rather than deleted, so `curl` still works while setting a model up without putting an inference API on the LAN. Verified against the running container: `127.0.0.1:8999/health` → 200, `192.168.20.4:8999` → refused. The comment now says the line can be deleted outright.

**3 — arm64 (medium).** Confirmed, and both halves of the old comment were wrong. The registry returns 404 for `cpu-arm64-1.9`, and `cpu-1.9`'s OCI index carries a single `linux/amd64` manifest — so it does not "run anywhere" either. The arm64 line is published unversioned as `cpu-arm64-latest`; the comment and the docs now say that, pinning caveat included.

**4 — lists naming four SDKs (medium).** Confirmed. Rather than adding `docling` to each string, the lists are derived from the predicates through `sdksWhere`, as `ModellessOCRSDKs` already was, and tested against the predicates instead of against wording. `TestCanOCR` covers `docling`, and a new case asserts every valid SDK serves at least one binding.

On the two you set aside as product choices: the CPU overlay's commented-out `app:` block is still commented out, and a document is still marked complete while embed runs — though the second is now only cosmetic, since nothing keys a decision on it.

Backend `build`/`vet`/`test -tags vectors` clean, 174 frontend tests, `vitepress build` clean.

PR #51, opened 2026-09-05
Make recent OpenAI models work: reasoning_effort, and the Responses API

### LINKED ISSUE #49: Chat is broken with recent OpenAI models.
```
openai search completion: POST "https://api.openai.com/v1/chat/completions": 400 Bad Request {
    "message": "Function tools with reasoning_effort are not supported for gpt-5.6-luna in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'.",
    "type": "invalid_request_error",
    "param": "reasoning_effort",
    "code": null
  }
```

A possible fix is here:
https://github.com/buldezir/lemmary/commit/7991eabf90ba1079e7d74eb7ad9fcadac31919f4

buldezir:
Closes #49.

The issue turned out to be two unrelated failures against recent OpenAI models. Both are fixed here, one commit each.

## 1. `reasoning_effort` refused alongside function tools

The error reported in #49, from `api.openai.com`:

```
Function tools with reasoning_effort are not supported for gpt-5.6-luna in
/v1/chat/completions. To use function tools, use /v1/responses or set
reasoning_effort to 'none'.
```

We never send `reasoning_effort` — these models default it server-side and then refuse the request because tools are present. The error names what would work, so `CompleteChat` takes the offer: pin `reasoning_effort` to `"none"` and send again. Remembered per model, because the Search loop runs to `maxSearchToolRounds` and the Research loop has no round cap, so without that every round pays a doomed request. Gated on the request actually carrying tools — pinning `"none"` on an extraction or OCR call would drop the model's reasoning where nothing asked us to.

Builds on @obestmann's [`7991eab`](https://github.com/buldezir/lemmary/commit/7991eabf90ba1079e7d74eb7ad9fcadac31919f4), with two corrections: the detector treats `param == "reasoning_effort"` as a positive signal rather than a gate (OpenAI-compatible proxies routinely omit `param`), and it reads `apiErr.Message` rather than `err.Error()`, which dereferences the `Response` and panics on an error value that did not come off the wire.

## 2. Some models are served only by the Responses API

Selecting `gpt-5.6-luna` broke everything it touched with a bare 500 and no upstream detail. Verified against the live endpoint — not something we sent:

| request | result |
| --- | --- |
| `gpt-5.6-luna` → `/chat/completions` | 500 — no tools, no `reasoning_effort`, no streaming, one-word prompt |
| `gpt-5.6-luna` → `/responses` | 200, tools included |
| `deepseek-v4-flash` → `/chat/completions` | 200, same body |

OpenCode Zen [documents a per-model endpoint](https://github.com/anomalyco/opencode/blob/main/packages/web/src/content/docs/go.mdx#L256) and routes `gpt-5.6-luna`, `grok-4.6` and the `muse-spark` models to `/responses`. Those are exactly the models our picker offered and could not call: luna 500, `grok-4.6` 401, `muse-spark` 403.

So the archive speaks both now. Call sites keep building the chat-completions request they already build; `responses.go` translates it and folds the `Response` back into a `*openai.ChatCompletion`. **No call site changed.**

The shapes disagree mainly about tool calls — chat carries them on the assistant message and answers with a tool-role message, Responses makes each a free-standing `function_call` / `function_call_output` tied together by `call_id`. Also handled: JSON mode → `text.format`, function tools flattened, `input_image` / `input_file` for LLM OCR, `store: false`, and reasoning items skipped on the way back.

Which endpoint a model lives on is **discovered, not listed** — a model string is whatever the operator configured against whatever gateway they point at, and a hardcoded list would be stale by the next launch. Try `/chat/completions`; on a status that means *this endpoint does not serve this model* (401, 403, 404, 405, 500, 501) translate and try `/responses` once. 502/503/504 are deliberately excluded — ordinary transient failures, where rerouting would be guessing. A model is remembered only when the fallback succeeds, and when the guess was wrong the caller hears the original error, not a 404 from an endpoint the provider never had. Streaming falls back the same way, but only while nothing has reached the reader.

## Verification

`GOFLAGS=-tags=vectors CGO_ENABLED=1 go test ./...` green across the backend.

New unit tests cover the fallback-and-remember cycle through a full `Research` loop, the translated request body item by item, the response converter, streaming, and the negative cases (a transient error does not reroute; a failed fallback keeps the original error and is not remembered). Mutation-checked: disabling the fallback fails 7 tests, corrupting `call_id` fails the shape test.

`responses_live_test.go` checks what a fake server cannot — that the translated request is one a real provider accepts. Skipped unless `LIVE_AI_KEY` / `LIVE_AI_BASE_URL` / `LIVE_AI_MODEL` are set. Against live `gpt-5.6-luna`:

```
POST .../chat/completions model=gpt-5.6-luna purpose=research round=0
WARN chat completions refused this model; retrying on the Responses API ... 500
POST .../responses model=gpt-5.6-luna purpose=research round=0 api=responses
POST .../responses ... round=1        <- no further 500s
... prompt_tokens=1154 cached_tokens=1056   <- prompt caching intact
reply: You paid **€412.90** for car insurance ... [Allianz car insurance 2026](/document/doc1)
```

`docs/ai_providers.md` gains a troubleshooting entry. Note `backend/internal/appapi/documents_timeline.go` already fails `gofmt -l` on `main`; left alone.

### DIFF
--- a/backend/internal/ai/modelnotes.go
+++ b/backend/internal/ai/modelnotes.go
@@ -1,0 +1,105 @@
+package ai
+
+import (
+	"strings"
+	"sync"
+)
+
+// What this process has learned about a model while talking to it. A provider
+// refusing a request the same way twice is worth remembering: the discovery
+// then costs a rejected request per model per process rather than one per call.
+//
+// Every note is keyed by endpoint as well as model. A model string means
+// nothing on its own -- an instance can bind several providers at once, each
+// with its own base URL (see internal/appapi/providers.go), and "gpt-5.6-luna"
+// behind OpenCode Zen is not the same endpoint as the same name at
+// api.openai.com. Keying on the name alone would let one gateway's quirk
+// follow the model onto another.
+type modelNote struct {
+	baseURL string
+	model   string
+}
+
+func noteFor(baseURL, model string) modelNote {
+	return modelNote{
+		baseURL: strings.ToLower(strings.TrimRight(strings.TrimSpace(baseURL), "/")),
+		model:   strings.ToLower(strings.TrimSpace(model)),
+	}
+}
+
+func (n modelNote) empty() bool { return n.model == "" }
+
+var (
+	// Models that would not take function tools alongside their default
+	// reasoning_effort, and had to be pinned to "none".
+	noReasoningEffortNotes sync.Map
+	// Models this endpoint does not serve on /chat/completions at all.
+	responsesAPINotes sync.Map
+	// Models /chat/completions has answered at least once, which makes a later
+	// refusal a transient failure rather than a wrong endpoint.
+	chatCompletionsWorked sync.Map
+	// How many times /chat/completions has refused a model with a status that
+	// could mean either.
+	ambiguousRefusals sync.Map
+)
+
+func rememberNoReasoningEffort(baseURL, model string) { store(&noReasoningEffortNotes, baseURL, model) }
+func needsNoReasoningEffort(baseURL, model string) bool {
+	return loaded(&noReasoningEffortNotes, baseURL, model)
+}
+
+func rememberResponsesAPI(baseURL, model string) { store(&responsesAPINotes, baseURL, model) }
+func needsResponsesAPI(baseURL, model string) bool {
+	return loaded(&responsesAPINotes, baseURL, model)
+}
+
+// rememberChatCompletionsWorked records a completion this endpoint served, so a
+// later failure is read as the provider having a bad minute rather than as a
+// model that lives somewhere else.
+func rememberChatCompletionsWorked(baseURL, model string) {
+	store(&chatCompletionsWorked, baseURL, model)
+}
+func chatCompletionsHasWorked(baseURL, model string) bool {
+	return loaded(&chatCompletionsWorked, baseURL, model)
+}
+
+// countAmbiguousRefusal returns how many times this endpoint has now refused
+// the model with a status that does not say which of the two it means.
+func countAmbiguousRefusal(baseURL, model string) int {
+	note := noteFor(baseURL, model)
+	if note.empty() {
+		return 0
+	}
+	count, _ := ambiguousRefusals.Load(note)
+	n, _ := count.(int)
+	n++
+	ambiguousRefusals.Store(note, n)
+	return n
+}
+
+func store(m *sync.Map, baseURL, model string) {
+	if note := noteFor(baseURL, model); !note.empty() {
+		m.Store(note, struct{}{})
+	}
+}
+
+func loaded(m *sync.Map, baseURL, model string) bool {
+	note := noteFor(baseURL, model)
+	if note.empty() {
+		return false
+	}
+	_, ok := m.Load(note)
+	return ok
+}
+
+// resetModelNotes clears everything this process has learned. Tests only.
+func resetModelNotes() {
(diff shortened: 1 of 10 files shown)


buldezir:
Findings 1–3 fixed in 21ff1ec. I reproduced each one first rather than taking them on faith; all three were real.

**1. The #49 400 never reached `/responses`** — correct, and it was the worse of the two options. `/responses` keeps the tools *and* the reasoning; `none` keeps the tools by switching the reasoning off, and we were remembering it for every later Search/Research round. That 400 now tries `/responses` first and settles for `none` only where there is no Responses endpoint to take it to.

The sub-point was a real bug, reproduced: `rememberNoReasoningEffort` ran before the retry came back, so a `none` the provider also rejected was still pinned. It is recorded on success now, matching `rememberResponsesAPI`.

**2. A single 500 rerouting a dual-API model** — you were right that I drew the transient line in the wrong place: excluding 502/503/504 while keeping 500, which is the most common one of the lot. Now:

- an endpoint that has **already answered** for a model is never rerouted, whatever a later failure says — that alone kills the `gpt-4o` case;
- an unambiguous status (401/403/404/405/501) is believed the first time;
- a 500 is acted on immediately but only *believed* the second time. A provider's bad minute does not repeat; a wrong endpoint does. Either way the caller still gets its answer — what changes is what gets written down.

The keying complaint was a separate bug I had missed, and is fixed independently: notes are keyed by `(baseURL, model)`. Providers are records with their own `base_url` and an instance can bind several at once, so `gpt-5.6-luna` at Zen and the same name at `api.openai.com` were sharing one verdict.

**3. `status: failed` treated as success** — reproduced, both halves:

```
err=<nil>
SURFACED AS SUCCESS: choices=[{"finish_reason":"stop","message":{"content":"",...}}]

text="half " err=<nil>
STREAM ERROR EVENT SWALLOWED
```

Confirmed your reading of the cause: `ssestream` raises an event as an error only via `gjson.GetBytes(data, "error")`, and a Responses error event carries `code`/`message` at the top level with no nested object. Any status but `completed`/`incomplete`/`in_progress` is an error now, `type: error` and `response.failed` frames are raised, and a failed generation no longer counts as proof the endpoint serves the model.

**4. `store: false` without `reasoning.encrypted_content`** — the mechanism is real and I have not fixed it. But "documented-broken" is not what I measured: the live luna run goes three tool rounds to a correct cited answer with prompt caching hitting `cached_tokens=1056`. It is a quality-and-cost gap on reasoning models, not a break, and closing it means threading provider-specific state through the chat-shaped transcript — the thing the translator exists to avoid. It wants its own change; noted in the commit message rather than papered over.

All three tests you asked for are in, plus the ones the fixes implied: a proven model is not rerouted by a 500, an unambiguous status is believed at once (401/403/404), `status: failed` is not a successful fallback, a stream error event is not swallowed, `none` is not remembered when it also fails, `/responses` is preferred over `none` when both work, and discovery does not leak between two providers sharing a model name.

Mutation-checked — each fix reverted individually, and every one is caught:

| reverted | tests failing |
| --- | --- |
| model-only keys | 2 |
| 500 as a certain mismatch | 3 |
| no response-status check | 1 |
| ignore the stream error event | 1 |
| remember `none` before the result | 1 |
| skip `/responses` on the 400 | 1 |

Full backend suite green, and the live check against real `gpt-5.6-luna` still passes end to end.

PR #56, opened 2026-09-08
Give documents awaiting review a place to wait, and show what the pipeline is doing

### LINKED ISSUE #50: [Feature Request] "Unsorted" inbox
I uploaded 3 files, they were processed and visible on the index page. But what I'm lacking is the "Unsorted" section, where docs will stay until I review them and click "Save" or something like that, that will remove them from Unsorted.

Use case: I'm uploading docs occasionally for a few weeks, and then in the end of the month I have some time to review them and fix mistakes possibly made by AI during OCR. For example, on a doc with multiple dates, the "Document date" field might be parsed from the wrong section of the page, so review is a mandatory step for me.

buldezir:
Closes #50.

The issue asks for an "Unsorted" section: somewhere uploads sit until they have
been looked at. Working on it surfaced the larger problem behind it -- a
document disappeared into the pipeline after upload and the UI said almost
nothing about it, so a 429 from an OCR provider or a quietly failed embed step
could only be found by opening the document and reading a raw list of step
records. This gives documents a tray to wait in, and makes what happens to them
legible.

## The Inbox

The Inbox is everything the pipeline has not finished with (`processing_status
!= completed`): a document that never came out needs attention as much as one
that came out doubtful. No new column, no migration.

`Settings -> Always require review for new documents`, off by default. With it
on, everything the AI extracted metadata for waits, and nothing reaches
`completed` except by your saying so. The header offers the Inbox only where
that setting is on -- without it a document reaches `needs_review` only by
extracting doubtfully, and a permanent link to an almost-always-empty list is
worse than no link.

Mark reviewed on the detail page, on a card, or on a page of cards at once, plus
a confirmed Delete selected. The Inbox is its own page rather than the documents
list under a flag: the two share their data and their rows and nothing else --
no status dropdown, because the path decides it, and no timeline, search or
filters, because narrowing a tray only hides work still to do.

## What the pipeline is doing

`processing_jobs` gains an `error` field, written by `failJob` and cleared on
claim so a retry cannot show the previous attempt's message. It is written only
when no step recorded the failure, since a step failure already has its message
in `step_runs`.

`summarizeJob` turns a job into one line -- the failed step and its provider
message, a retry, a soft-failed embed on a document that otherwise reads
"completed", or a queue stalled on an unconfigured provider. It appears on the
cards, on the document page, and on the new `/activity`, which lists what is in
flight and what failed in the last day with a Reprocess button. Activity reads
`processing_jobs` directly, so there is no endpoint behind it.

## Decisions worth a reviewer's attention

**Review is required because the AI wrote the metadata, not because a document
is new.** The override lives in `apply_metadata` and deliberately not in
`finalizeDocumentWithoutApply`, the paperless-ngx preserve-mode import -- which
would empty a migrated archive into the Inbox for extraction that never ran.
`TestThePolicyCannotReachAStepListWithoutApplyMetadata` guards that.

**"Unfinished" is the absence of a status**, so it needed a name:
`models.StatusFilterUnfinished`. PocketBase negates; the search index cannot,
because a negated term also matches documents with no status indexed, so that
side takes the set. The header badge counts the same set the list shows.

**The Inbox's "no filters" rule is enforced in the hook, not only in
`validateSearch`** -- `useSearch({ strict: false })` returns raw params, so a
hand-typed `?q=` reached the query and emptied a tray with no search box to
explain it.

**`always_require_review` is on `/api/app/meta`**, because it decides what a
regular user's list URL means and `/api/app/settings` is admin-only to read.
Meta is now subscribable (`onAppMetaChanged`), or the nav keeps the old answer
until a reload.

**No new endpoint for marking reviewed or deleting** -- both are plain owner
writes the collection rules already permit.

## Fixed along the way

- `getActiveJobCounts` counted by status, but `apply_metadata` writes
  "completed" onto a job before `embed` runs: the badge said zero while Activity
  listed the work, and a stale-data sweep could start on a running pipeline.
  Both halves are now bounded by `finished_at`.
- Card job lines refetched only when the page's documents changed, so a line
  froze at "OCR - 12.0s" behind a document sitting at "processing".
- `CardDescription` claimed "Low extraction confidence" for any review-pending
  card without a summary, and read a never-recorded `0` as low.
- Requiring review also filtered a bare `/` down to Completed; that made the
  default something the URL layer had to know, so `/` is the whole list again.

## Known consequence

With the setting off, doubtful extracts are reachable through the Documents
status filter but not through the header, and not through Activity, whose jobs
have finished. Whether the link should follow the work instead is left open.


buldezir:
Review addressed — all three findings were real. Pushed 9bda69fd (plus 8513327 on the overlay branch).

**1. Inbox list ignored its own writes.** Confirmed: only the header badge listened to `documentEvents`, so with no realtime a cleared card sat on the Inbox while the count already said it had gone. The list now subscribes to the same event as the badge, which also covers any future local write. Locked by `a cleared card leaves the list even with no realtime`, which aborts `/api/realtime` so only the local path remains.

**2. Last page showed the success empty state.** Confirmed. The load now walks back to the last real page when the URL points past the end, instead of rendering an empty page that reads as an empty list. Locked by `clearing the last page walks back rather than claiming the Inbox is empty` (13 stubbed documents, so page 2 exists).

**3. Back link and post-delete went to `/`.** Confirmed, and the same class of bug as the upload landing this PR already fixed. The back link, its label and the post-delete redirect now follow one rule, so `landingAfterUpload` is `documentsLanding`.

Worth noting for the record: the reason the first pass of the browser specs could assert Mark reviewed *sent* a PATCH but not that the document *left the list* is that the stub answered every query with the same fixed set. It now serves a mutable Inbox that a PATCH removes from, so the list, the badge and the empty state are asserted against one source of truth — which is what made findings 1 and 2 testable at all.

`./scripts/test-all.sh` green: 216 frontend unit tests, 109 browser tests, the Go e2e suite, backend unit tests.

Separately, at the author's request this push also cuts the comments back to the parts that are genuinely hard to see — the ngx-import exemption, the job/document split, why both halves of the URL round-trip read the same default, why `reviewPolicy` is module state rather than a hook, and why the page clamp exists (−323 lines, +198).

PR #57, opened 2026-09-08
Show the document beside the fields it describes, instead of in another tab

### LINKED ISSUE #48: [Feature Request] File preview on the file page
Image

It'd be great if we could have file preview (on the left or right) on the document page, just like in Paperless NGX

buldezir:
Correcting a title or a date meant reading the file in one tab and typing in another. The detail page now frames the file itself beside the metadata form — a PDF in the browser's own viewer with page navigation and find-in-document, an image fitted to the pane — and a **Preview** button folds the column away when the fields want the width. The choice is remembered in `localStorage`.

Closes #48.

Document detail with the file beside its metadata

## Layout

Left of the form at `xl`, above it below that — but **after** it in the DOM. Stacking order is CSS; tab and screen-reader order is not, and a keyboard should reach the fields without walking through a PDF viewer's own controls first.

## Which files get a pane

Only pdf, jpeg, png and webp. docx and xlsx have nothing a browser can frame; txt and csv could be framed, but their text is already in the OCR-text field below, so framing them would say the same thing twice. Those documents get no column and no toggle rather than a viewport-tall box apologising for itself — they keep exactly the page they had.

`previewKind()` makes that call from the stored filename. The record carries no mime type, and PocketBase appends a content-sniffed extension when an upload arrives without one, so the extension follows the bytes rather than the uploader's claim.

## Why there is no blob layer

The file goes into the frame as a plain tokened URL, with nothing fetched through JavaScript. Worth saying, because the obvious reading of the server's behaviour says that cannot work: PocketBase stamps every file response with `default-src 'none'; media-src 'self'; style-src 'unsafe-inline'; sandbox`, and a fully sandboxed response sounds exactly like something that would stop a built-in PDF viewer.

Both Chrome and Firefox render it anyway. I checked directly, against that header, before writing anything. Believing the inference instead would have bought a blob fetch, a revoke lifecycle, an abort path and the whole file held in memory — to work around a restriction that is not there.

## Two bugs fixed at the root

Both in `fileUrlWithToken`, both reachable before this change:

- `pb.files.getToken()` derives its auto-cancel key from method and path, so two concurrent mints aborted each other and the loser rejected as an abort. One path, one key — double-clicking **Open file** was enough. Now opts out with `requestKey: null`, as the other concurrent callers in that file already do.
- `getURL` returns `""` for an empty filename, and fetching `""` resolves against the current page: `index.html`, with a 200, which no response check would catch. A document with no file now says so.

It also never called `ensureAuth()`, unlike every other authed helper there, so a tab left open past expiry took a 401 from the token endpoint.

## Verification

`./scripts/test-all.sh` passes every stage; `docker build` succeeds. Overlay branch: `worktree-document-preview` (**push it first** — the PR job falls back to `main` and looks green otherwise).

Three browser specs cover the wiring the app owns, which is narrower than "the document is readable" on purpose: headless Chromium ships without the PDF viewer, so a framed PDF is blank there whatever the server sent. They pin that the frame's `src` is the protected file URL with a token, that the server answers it, and that it is fetched **exactly once** even though the page re-renders about once a second while the pipeline runs — that count is the real guard on the pane depending on `id`/`collectionId`/`file` instead of the record object the poll replaces.

The screenshot generator had the same headless trap, and worse: it would have quietly published a documentation set with an empty box where every preview should be. It now launches the full Chromium build, headed, and the runbook says to wrap it in `xvfb-run` where there is no display. The four affected images are reshot from the demo library.


buldezir:
Thanks — two of the three were real, and the second one was the good catch. All addressed in d495e14c (specs in `lemmary-dev@24a41b7`, pushed first).

**Hiding unmounted the viewer** — agreed, and this was the worst of the three, because it broke the feature's own premise: take the width for the fields, give it back, and you are on page one at default zoom with find-in-document forgotten. The aside now stays mounted and goes `display:none`.

I checked that the fix is worth anything before relying on it, since it would be a no-op if browsers reset a hidden frame. Scrolled an 8-page PDF to page 8, hid it, showed it: still page 8. Remounted the same `src` as a control: back to page 1. It is also not mounted until the first show, so a reader who turns the pane off does not pay for the download.

**The mobile layout** — agreed, and stronger than "default the flag off": below `xl` the pane has no premise at all. "Beside the fields" needs two columns; narrower than that it was a 70vh block above the form, and on the browsers you name a 70vh *blank* one, with the fields the page exists to correct pushed under it — as the default. So the pane simply does not exist below `xl` now, and the page there is exactly what it was before this feature, `Open file` included. Gated on a watched media query rather than a CSS hide, precisely so a phone does not spend the download on something it will never show; widening the window brings it back without a reload.

**The token in `src`** — added `referrerPolicy="no-referrer"`, but I want to be accurate about why, because the stated risk is not quite the mechanism. Chrome's default is `strict-origin-when-cross-origin`, so a cross-origin navigation out of the framed document sends the origin only — no path, no query, no token. The header PocketBase sets does not loosen that. So this is defence in depth, not a leak being closed, and it is one attribute, so it stays.

The exposure worth naming is the other one: `Open file` has always put the same tokened URL in a visible address bar and browser history. The pane is the quieter end of it. Not changing that here — it is the iOS and print escape hatch — but it is where the token is actually on display.

**`onError` on the iframe** — declined, with a comment saying why. A refused frame gets an error *document*, which loads successfully as far as the element is concerned, so the handler never runs; `onError` fires for network-level failures, not a 403. PocketBase's message renders in the frame instead. The `` branch does get an `onError`, where it does fire.

**Error styling and the missing test** — both taken. Mint failures are `text-madder` now rather than sharing `text-ink-soft` with "Loading preview...", and `documents.test.ts` covers the empty-filename throw, which as you say needs no mock.

Four browser specs now, and the rewritten toggle test asserts the frame is hidden but still mounted with an unchanged `src` — a fresh token being exactly what a rebuild would produce. Both wide specs pin 1440×900, since the project default is 1280, sitting on the edge of the media query they depend on.

PR #61, opened 2026-09-10
Tell AI providers who is calling, except the one that vets its callers

buldezir:
Every outbound AI request identified itself as `OpenAI/Go 1.12.0` (SDK clients) or `Go-http-client/1.1` (hand-rolled ones). A provider reading its logs could not tell this app from any other Go program, and a gateway that rate-limits or blocks by agent had nothing of ours to name.

## What changed

One constant and one helper in `aiprovider`, used at every site that talks to a provider:

- **SDK clients** — `NewOpenAIClient` and `NewEmbedder` append `UserAgentOptions(sdk)`, covering chat, extraction, LLM-based OCR and embeddings.
- **Hand-rolled requests** — Mistral OCR, docling, and the models listing set the header directly.

## The ChatGPT exception

`SDKChatGPT` sends nothing new. The Codex backend serves only its own clients: it checks the `originator` header against a whitelist and refuses anything else with a 403, so an unfamiliar user agent is the next thing it would notice. That path keeps looking exactly like the SDK it impersonates.

The gate lives inside `UserAgentOptions` rather than at each call site. The models listing needs no gate of its own, since it answers the Codex catalogue locally before building a request, and ChatGPT device auth was left untouched.

No version in the string: nothing in the build stamps one, and an agent claiming a version that never moves is worse than one claiming none.

## Test

`TestChatUserAgent` drives a real client at an httptest server for three SDKs and asserts what each one hears: our name for OpenAI and OpenCode, the SDK's for ChatGPT.

Full `scripts/test-all.sh` passes. The first run hit a flaky browser e2e stage; the rerun was green across all seven stages.

### DIFF
--- a/backend/internal/ai/embed.go
+++ b/backend/internal/ai/embed.go
@@ -95,6 +95,7 @@ func NewEmbedder(sdk, apiKey, model, baseURL string, dims int, timeout time.Dura
 	if sdk == aiprovider.SDKOpenCode {
 		opts = append(opts, option.WithMiddleware(aiprovider.SessionMiddleware()))
 	}
+	opts = append(opts, aiprovider.UserAgentOptions(sdk)...)
 	// A keyless provider sends no Authorization header at all, rather than an
 	// empty "Bearer ". The local SDK is the case: a sidecar on the compose
 	// network has nobody to authenticate to, and an endpoint that does read the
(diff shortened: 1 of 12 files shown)


buldezir:
Both points were right, fixed in 4b49420.

**The gap.** `opencode.NewMessages` builds an Anthropic client for the third of the OpenCode catalogue served on `/messages`, and the openai-go option never reached it. Those requests were going out as `Anthropic/Go 1.71.0`. It now sets the header unconditionally, since that client is only ever built for an OpenCode base URL and the Codex exception cannot apply to it.

**The assertions.** Added where a harness already existed: the messages client beside the credential headers it is pinned on, the models listing, and the embedder. Removing the header from `NewMessages` fails the first one with the old `Anthropic/Go 1.71.0` value, so it bites.

Mistral OCR is still unasserted. It has no test server at all, and standing one up costs more than the single line it would guard. Left it out deliberately rather than by oversight.

Full `scripts/test-all.sh` green.

PR #62, opened 2026-09-10
Serve OpenTelemetry metrics on a port of their own, so a scraper holds no key

buldezir:
`/api/health` and the `_logs` table were the whole observability story. This adds a Prometheus scrape endpoint on its own port, plus instrumentation of the request path, the processing pipeline, every outbound AI/OCR call, and how full the instance is against its `LIMIT_*` allowances.

Off unless `METRICS_ADDR` is set, and off is exactly the old behaviour.

## Why a separate port

A route would put telemetry behind the archive's own auth — or worse, in front of it. On its own port, whatever scrapes it needs no credential for the app at all.

`METRICS_ADDR=9464` or `:9464` binds every interface, which is the only form a container reached through a published port can use. The mapping is deliberately **not** in `docker-compose.yml`: an unauthenticated endpoint should not become internet-reachable because somebody set a variable. Spell out a host (`127.0.0.1:9464`) to keep it on loopback on bare metal.

## What it exposes

```
http_server_request_duration_seconds{http_request_method,http_response_status_code,http_route}
lemmary_job_duration_seconds{outcome}                  # completed | needs_review | failed | retry
lemmary_jobs_pending
lemmary_ai_call_duration_seconds{kind,sdk,model,outcome}   # kind: chat | embed | ocr
lemmary_ai_tokens_total{model,kind}                    # kind: prompt | cached | completion
lemmary_usage{resource}          lemmary_limit{resource}
lemmary_usage_bytes{resource}    lemmary_limit_bytes{resource}
go_*  process_*
```

Nothing is labelled with a path, a document id, or an account.

## Instrumentation sites, and why these

- **Requests** — otelhttp's own middleware, adapted by the `apis.WrapStdMiddleware` that already existed, bound one priority inside vault's inflight wrapper. PocketBase's router fills in `http.route`, so latency is per route and not per URL.
- **Jobs** — `worker.runJob`, after the claim. The outcome comes from re-reading the job's status, **not** from the error: `handleStepFailure` returns `nil` when it re-pends for another attempt, so a nil error covers both "done" and "will try again" and those are not the same measurement.
- **AI and OCR** — four logical wrappers (`Complete`, `completeStreaming`, `embedBatch`, and one decorator over the `ocr.Provider` interface). Not a shared `http.RoundTripper`: there isn't one, and Google Vision goes out over gRPC where no HTTP hook could ever see it. One measurement per call the caller made, retries and endpoint fallbacks included.
- **Tokens** — `ai.logUsage`, which all twelve usage sites funnel through.
- **Usage vs limits** — `limits.Measure`, the same reading the Settings usage endpoint already serves, so a dashboard and the app cannot disagree about how many documents there are.

## Two decisions worth a look

**An unset `LIMIT_*` emits no series at all.** `0` is an allowance somebody sells — `LIMIT_ADDITIONAL_USERS=0` is "this account and no others" — so it cannot double as "unbounded", and `-1`/`+Inf` would be a magic number every query had to know about. Utilisation is `lemmary_usage / lemmary_limit`, absence means no ceiling. (`ocr_pages` is published too, though not a `LIMIT_` var: it's the cap that binds on every install.)

**The OCR decorator needs two types, not one.** Embedding `ocr.Provider` alone would hide `MaxConcurrency` from the type assertion in `pdfsplit` and quietly aim a page per core at a sidecar that asked for one request at a time. A second type is smaller than that bug, and a test pins both directions.

## Series count

otelhttp's defaults are sized for a fleet behind a load balancer: three 19-bucket histograms per route, five of whose labels are identical on every series one instance will ever emit. Two SDK views trim it — body sizes dropped, listener-describing attributes filtered — and `WithoutScopeInfo` drops the `otel_scope_*` labels, since there is one scope here. Measured on a live server: **84 series** instead of ~1500 for two requests.

No instrumentation package for Go runtime metrics either — the otel Prometheus exporter registers into client_golang's default registry, which already carries the Go and process collectors. `metrics_test.go` asserts `go_goroutines` and `process_open_fds` are on the page, which is what fails loudly if that stops holding.

## Dependencies

Versions are set by the exporter: `exporters/prometheus@v0.68.0` requires `otel v1.46.0`, so otel and otelhttp move up with it. otel was already in the module graph as an indirect dep via `google.golang.org/api`.

A side effect worth naming: setting the global `MeterProvider` switches on the instrumentation already inside the Google Vision client (`otelgrpc`), for free.

## Verification

- `./scripts/test-all.sh` — all stages pass (unit, API e2e, 141 browser e2e, frontend unit/build, docs build). No overlay change needed: the endpoint is off unless `METRICS_ADDR` is set, so no e2e behaviour moves.
- `docker build -t lemmary:local .` — passes (`go.mod` changed).
- Live server: endpoint binds `*:9464`; app port 404s `/metrics`; `lemmary_usage{resource="additional_users"}` moved 0 → 1 when a second account was created; an explicit `LIMIT_ADDITIONAL_USERS=0` published `0` while unset limits published nothing; the port closes on a clean stop.

New tests: `internal/metrics/metrics_test.go` (scrape shape, address parsing, off-when-unset), `internal/appwire/usagemetrics_test.go` (the zero-vs-unset rule), `internal/worker/job_outcome_test.go` (the retry outcome), and one in `internal/ocr` for the concurrency-limit passthrough.


buldezir:
All three findings confirmed and fixed in 2c5b9f0e. Each was reproduced before changing anything.

**1. Request statuses (High) — real, and verified live.** `tools/router/router.go:131-151` calls `ErrorHandler(resp, req, err)` after the hook chain unwinds, so a wrapper bound into that chain sees an unwritten response. Before the fix, two requests that genuinely answered 401 were both recorded as `status_code="200"`.

Wrapping `e.Server.Handler` outside the mux fixes the status but loses `http.route`, since `ServeMux` sets the pattern on the request it passes inward. What a hook on the chain *does* have is the error, so `appwire.recordRequest` derives the status with `router.ToApiError` — the same mapping `ErrorHandler` is about to apply — falling back to the written status, then to 200. Both attributes are now correct:

```
before:  401 -> "200"   401 -> "200"
after:   200 -> "200"   401 -> "401"   302 -> "302"   (http_route intact)
```

That drops otelhttp to an indirect dep and removes its two trimming views with it — the histogram is ours now, so attributes are chosen rather than filtered. Kept the semconv name and attribute keys; bucket boundaries are spelled out because the SDK defaults are milliseconds.

**2. Queue gauge (Medium) — real.** `QueueDepth` now takes `func() (int64, error)` and returns the error, matching `RegisterUsage`.

**3. DefaultRegisterer — real and reproducible.** `go test -count=2` served a page with none of the lemmary metrics. Now a private registry with the Go and process collectors registered explicitly, and `net.Listen` moved ahead of everything global so a failed bind leaves nothing behind.

Reproducing that surfaced a layer underneath: a private registry alone doesn't fix `-count=2`, because OpenTelemetry takes a single global `MeterProvider` — the second install leaves every instrument feeding the first while a second endpoint serves an empty page. `start()` now returns the endpoint already running, which is also the right answer for the e2e harness, since it boots a whole app repeatedly in one test binary.

New tests: `TestResponseStatus` (401/404/400, non-ApiError, written-status, error-outranks-written), `TestRouteOf`, and the scrape test asserts a 401 records as 401. `./scripts/test-all.sh` all green; `-count=2` passes; live server shows 93 series with `go_*` and `process_open_fds` present and the port closing cleanly.

PR #63, opened 2026-09-10
Let the admin add house rules to the extraction prompt, and give Settings tabs

buldezir:
## Extra extraction rules

The extraction system prompt was a fixed string literal, and the only part of it an admin could vary was the result language. Everything else — how to pick a title, which correspondent to prefer, what to tag — was baked in, so an operator who wanted *"treat Rechnung as the document type Invoice"* had to fork.

`app_settings` gains **`extraction_rules`**: up to 4000 characters, appended to the built-in prompt. Empty (the default) leaves the prompt byte-for-byte what it was.

It is an addition, not a replacement — the prompt's field list is coupled to `models.ExtractedMetadata`, and the catalog blocks are what make correspondent/type reuse and dedup work, so both stay. The rules land after those catalog blocks and **before** the closing `document_date` / no-markdown block: the format instructions a model follows best are the ones it read last, and a house rule is not entitled to loosen the JSON contract. The block says so, and unlike the catalog it is not labelled untrusted — only an admin can write it, through a route behind `bindAdmin`.

Tenant-owned, like `processing_result_language` and `always_require_review`: out of `touchesManaged()`, so a managed instance keeps it. No env var — this is a Settings choice, not a boot secret. `ExtractMetadata` logs the length as `rule_chars`, so "why did this document come out like that" has an answer in the log without the operator's own text ending up there.

`NewExtractor` took the new argument; `NewOpenAIClient` did not. That constructor is shared with chat, search, the splitter and LLM OCR — five callers and ~20 test call sites that already pass two empty strings for extraction-only fields — and `NewExtractor` has exactly one caller, so it sets the field on the client it builds.

## Settings tabs

`routes/settings.tsx` was 966 lines rendering six sections stacked on one screen, sharing one `<form id="app-settings">` with five Save buttons attached to it by id — each of which saved every field on the page. One more field made that worse, not better.

Settings is now a shell with one tab per section, on real paths, which is what `/upload` and `/import` already do:

| Path | Tab |
| --- | --- |
| `/settings` | Appearance |
| `/settings/ai` | Providers + Models |
| `/settings/processing` | timeouts, languages, extraction rules, review policy |
| `/settings/worker` | job timeout, retries |
| `/settings/duplicates` | detection, threshold |

Appearance sits on `/settings` itself, the way Files sits on `/upload`, so existing links to `/settings` still land somewhere sensible.

A form can no longer span sibling routes — and the AI tab has a provider form of its own, which would have nested — so each tab owns its form and PATCHes only its own fields. Every field in `settingsPatchRequest` is already a pointer where nil means "keep what is stored", so this needed no API change, and it deleted two things: `FORM_ID` with its scattered `form={FORM_ID}` attributes, and the `...(aiEditable ? {...} : {})` spread that kept a managed instance from naming fields it would be refused.

**Behaviour change:** Save under Processing no longer writes your edits on the Worker tab. That is the readable behaviour; the old Save quietly wrote five sections at once.

Shared plumbing is one hook, `useSettingsForm`, on the existing `useAsync` — no React context, of which this codebase has none. The providers list and drawer moved to `components/settings/ProvidersBlock.tsx`, so the largest file here is now ~280 lines. `tabClassName` was already copy-pasted in `upload.tsx` and `import.tsx`; a third copy is where it belongs in `components/ui.tsx`, and all three import it.

## Always require review

Moved to the top of the Processing tab, in a bordered panel with a larger accented checkbox. It is the only setting there that decides where a finished document goes, and as small muted text at the foot of a two-column grid it was the easiest thing on the page to miss.

## Verification

`./scripts/test-all.sh` — all stages pass (143 browser tests, API e2e, unit tests, frontend/docs builds).

E2E for both halves is on the matching **`lemmary-dev`** branch `worktree-extraction-prompt-customize`, pushed first: tab-aware `openSettings`, the layout assertions rewritten off the single page, three new specs (each tab reachable by URL with one Save; a tab's PATCH carries only its own fields; the rules field round-trips), and `TestAppSettingsExtractionRules`.


buldezir:
Both fixed in d5cf4d44.

**1. AI / Duplicates tabs lying while meta is in flight** — a real bug, and the distinction you drew is the right one: `aiManaged !== false` is correct for deciding what to *offer* (unknown and failed both have to read as managed, or a Save sends fields the server refuses), but these tabs go further and *state* who owns the settings.

`useAppMeta` now also answers `metaLoaded`, separating "not answered yet" from the two answers that both read as managed. `getAppMeta` resolves with its fallback rather than rejecting, so `data !== null` covers a failed request too — it answered, and the answer is unknown. The two tabs wait on that before naming an owner; the `aiManaged !== false` decision is unchanged, so a failed fetch still gets the notice rather than the fields.

No test for it: the frontend has no jsdom or testing-library, and a Playwright spec cannot reliably observe a single render frame.

**2. Per-document prompt attribution** — this was on my deferred list, and you're right that it shouldn't have been. The point isn't that the record is incomplete, it's that it became wrong: `extraction_prompt_version` exists to trace metadata back to a prompt, and rules change the prompt while the version stays `v1`, so a run recorded under it names a prompt nobody can reconstruct.

`ai.ExtractionPromptFingerprint` appends a short digest of the rules — `v1` with none set, `v1+rules.<6 hex>` with any. Enough to tell one rule set from another in the step tooltip, which is all it is for; not a checksum anyone verifies. No rules means the bare version, so existing runs and every instance that sets none read exactly as before, and `PromptVersion` is a string in the `step_runs` blob, so no migration.

**3. The untested wiring** — closed in-tree, since you're right that overlay-only coverage leaves each of these one omitted line from a silent miss:

- `migrations`: the round-trip test now also loads the config off the app — the record is only worth anything if the settings the extractor is built from carry it.
- `appapi`: `settingsResponseFromConfig` hands the rules back, so a stored value can't be one the Settings page never shows.
- `worker`: a new `step_runs_test.go` pins the fingerprint onto the extraction step, beside the embedding step's existing test for the same function.

Docs no longer describe `EXTRACTION_PROMPT_VERSION` as the whole of what gets recorded. `./scripts/test-all.sh` passes: 143 browser tests, API e2e, unit tests, frontend and docs builds.

PR #72, opened 2026-09-13
Make tags a vocabulary the user curates, not one the model invents

buldezir:
## Why

Extraction created a tag record for every name the model answered with, and `mergeTagNames` doubled that by unioning `tags` with `tags_translated` — a German invoice left behind both `Rechnung` and `Invoice`. Nothing ever offered the model the tags that already existed (`ExtractionCatalog` carried correspondents and document types only), so it re-invented vocabulary on every document and the tag table drifted into hundreds of near-duplicates nobody chose.

Tags are now a closed set the user writes by hand. The model is given that set and assigns from it; it no longer adds to it.

## What

**Prompt** — `ExtractionCatalog` gains `Tags`, and `formatAllowedTagsPrompt` appends it beside the other two catalogs. It deliberately does not reuse `formatExistingNamedListPrompt`: that one ends *"only invent a new X when none of these match"*, which is exactly what tags must not do. The untrusted-data framing is kept. An empty vocabulary asks for an empty array. `tags_translated` is gone from the contract and from `ExtractedMetadata` — a vocabulary its owner wrote has nothing to translate.

**Apply** — `ensureTags` becomes `matchTags`, which resolves names against the user's existing tags and returns the ones it could not, for the log. The match is normalized (`normalizeNamedEntityKey`, shared with correspondents), so a model answering `invoices` for a tag named `Invoices` still lands rather than losing a tag over a capital letter. The user's tags are read once per document, not once per name.

`EnsureTag` stays exported and unchanged: archive restore, the paperless-ngx import and the ngx REST API still create tags, because those move data the user already owns.

**Prune** — tags are exempt. Correspondents and document types are extraction output, so an unreferenced one is debris; a tag is unreferenced *precisely between being created and first being used*, and pruning them would delete the vocabulary its owner had just typed. `PruneResult.Tags` stays in the shape, always `0`, because the API response and the Management page read it.

**UI** — a Tags page under the cog menu (not admin-only; tags are per-user) lists, creates, renames and deletes them through the PocketBase SDK. The collection's rules have been user-scoped since tags became per-user, so no Go endpoint was needed. The unique-index violation is surfaced as *"You already have a tag called X"* rather than *"Value must be unique."*

The document editor's "Tags (comma separated)" box — the frontend's own second creation path — becomes chips plus a `Combobox` over the rest of the vocabulary, and `saveDocumentMetadata` takes ids instead of names. An empty vocabulary links to `/tags` instead of offering an empty picker.

Copy that claimed the old behaviour is corrected in the result-language hint, Management's prune block, and `docs/{ai_providers,comparison,screenshots,setup}`.

## Tests

New: `matchTags` unit tests (exact hit, normalized hit, unknown dropped, another user's tag not matched, creates nothing), extraction-prompt tests for the closed vocabulary and the absent `tags_translated`, and a `taxonomy` prune test — that package had none.

Overlay branch `t3code/manual-tags-system-prompt` (pushed first) carries the e2e half: an API test asserting a processing run leaves the vocabulary untouched and matches `plumbing` against a user-written `Plumbing`, a browser spec for the page and for the editor's "cannot create" contract, and updates to the three suites that asserted the old behaviour.

`./scripts/test-all.sh` — all stages pass.

## Not doing

No migration: the schema is unchanged and existing AI-created tags become the starting vocabulary. No tag filter on the document list (the agent's `tags` argument and the ngx API already filter). No `name_original` on tags.


buldezir:
Thanks — findings 1, 2 and 4 were real; fixed in 505bb78d. Detail on each, including the two I handled differently than suggested.

**1. Tag catalog failure clearing tags — fixed, and it was worse than "a hole".** You're right that the comment I left there talked myself into it. `loadExtractionCatalog` now returns an error and `ExtractMetadataStep` fails when the tag list can't be read. Correspondents and document types still warn-and-continue, because those genuinely are advisory — a model that never sees them still answers something usable — whereas an empty tag list is an instruction to return nothing, which apply then writes over the document. A document with no owner now fails for the same reason instead of being extracted against an empty vocabulary.

Related, from your "also noted": `matchTags` with an empty user id returned `[]` with no error where `EnsureTag` used to error. Same failure mode — a no-op that clears rather than fails. It errors now.

**2. Editor hiding tags — fixed, and the false comment with it.** Chips no longer wait on `listTags`: they render from the document's own `expand.tags`, which the list cards already use, so a slow or failed vocabulary load can't blank a row the cards still populate. The vocabulary is only needed to offer the *rest*, and its error is now shown rather than swallowed.

You were right that "the list as shown" was false — save writes `tagIds`. Rather than make the display authoritative, I made it honest: an id with no name behind it renders as "Deleted tag" instead of vanishing, since it would have been written back either way, and now it can be removed deliberately.

**3. The 500 cap — logged, not lifted.** Agreed the prompt claims completeness it can't guarantee; I disagree on softening the prompt, which is what "not claiming completeness" would mean in practice. Any hedge there re-opens the invention this PR exists to close, and the truncated case is rare and fixable by the user, while invention was the default and wasn't. So the cap stays and the overflow is logged instead — "some tags are never assigned" is otherwise a symptom with no trace anywhere. The extraction start line also carries `catalog_tag_names` now, which is your last "also noted" item.

**4. Same-tab link — fixed.** `target="_blank" rel="noopener noreferrer"`. It only renders mid-edit, so it was throwing away half-corrected metadata.

**Not changed**

- *Catalog dedupe `ToLower` vs matching `normalizeNamedEntityKey`.* Correct observation, benign outcome: two punctuation-variant tags both reach the prompt and collapse to one id on apply. The document ends up with the right tag, once. Not worth a second normalization pass on the prompt side.
- *Frontend unit tests for `saveDocumentMetadata({ tagIds })` and `TagField`.* There is no `vi.mock` anywhere in `frontend/src` — that layer is pure functions only, and mocking the PocketBase SDK to cover two thin wrappers would introduce a pattern the repo doesn't have. I exported and tested `duplicateNameError`, which is the one actual decision in `tags.ts`. The rest is covered in the overlay's `browser/tags.spec.ts`, which drives create/rename/delete, the duplicate-name message, and the editor's "offers the vocabulary, cannot create" contract end to end.
- *An e2e for finding 1.* Forcing a DB error on the tags table isn't reachable from the harness; the contract is pinned by `TestLoadExtractionCatalogEmptyUser` and `TestMatchTagsRequiresAUser` instead.

`./scripts/test-all.sh` — all stages pass.

PR #73, opened 2026-09-14
Let a half-typed word find the word it starts

buldezir:
## Summary

- Match terms of three or more letters against indexed word prefixes.
- Rank exact whole-word matches above prefix-only matches while preserving AND semantics.
- Document prefix behavior and expand full-text coverage tests.

## Testing

- Added unit tests for prefix matching, ranking, short prefixes, and multi-term queries.
- Updated search evaluation thresholds and relaxed-search assertions.
- `./scripts/test-all.sh` — Not run.


buldezir:
All four findings addressed in 7a0cee32; `./scripts/test-all.sh` green.

**1 (High) — confirmed, and worse than "it might be slow".** Benchmarked against an OCR-sized vocabulary (120k terms, 20k sharing one three-letter head, 3k documents): one three-letter term cost **415ms**, against 3ms without the prefix leg.

Two narrower fixes were measured and rejected before the one that landed:
- Skipping `ocr_text` kills the cost *and* the benefit — strict recall@5 falls straight back to 0.511, because the compounds and inflections a prefix rescues live in the OCR text, not the metadata.
- A longer floor for `ocr_text` is not a bound: a dense six-letter prefix still cost 155ms. Length is a proxy for expansion, not a limit on it.

So prefix legs moved to a **widening rung** — the ladder this file already uses for relaxed queries. The strict path now has a fallback it never had: same conjunction, each term widened to a prefix, run only when the whole-word query matched nothing. Queries that already match do no prefix work at all (0-2ms in that benchmark). Retrieval numbers are unchanged: strict recall@5 0.598, MRR 0.609.

Remaining ceiling, marked `ponytail:` in the code: an empty-result query still pays one wide expansion. The upgrade path is a `FieldDictPrefix` size check, if a large archive ever makes typing through a dense prefix slow.

**2 (Medium) — correct, and it dissolves.** A prefix hit can no longer appear beside an exact one, because the rung producing it only runs when there are none, so the title-prefix-4×0.5-beats-exact-OCR-1.0 case cannot arise. `prefixBoost` is 0.2 so exact stays ahead within the rung too. Docs no longer claim "always outranks" — they say when the retry runs. New test covers title-prefix vs OCR-exact.

**3 (Medium) — fixed.** `docs/paperless_ngx.md` now describes the prefix rung and its limits instead of promising whole-word-only.

**4 (Low) — taken.** `prefixTerm` rejects digit-bearing terms, sharing `fuzzyWorthy`'s rule as `hasDigit`. It also takes the leading token now: `amaz,` was being searched for literally.

**Tests** — added the three-rune floor, a trailing comma, a German compound in OCR text, digits, exact-vs-prefix across fields, and counts agreeing with the list. `CountMatching`/`MatchingIDs` escalate the same way, or a grouped count would report zero for the page in front of the user. The eval's exact-class check now asserts strict finds each document outright, rather than only failing when relaxed rescued it.

**`chunkTextQuery`** — left alone deliberately: passage search takes model-written keywords, not a half-typed box, and is already hybridised with vectors. "Every caller" in the earlier commit message meant every caller of `fieldQuery`; that was overstated and is corrected in the new one.

Also merged `main`: the overlay browser suite had moved to the timeline folding behaviour from #74, which this branch predated. Unrelated to the search change.

PR #82, opened 2026-09-14
Let Deep Research and Ask AI reach the web, when asked twice

buldezir:
Both surfaces could only reason over documents the archive holds, so a question whose answer moved on — a rate that changed, a company's present details, a number never filed — came back as *the archive does not contain this*, or out of the model's memory.

Adds `web_search` and `web_fetch`, backed by a new `websearch` package speaking Tavily's `/search` and `/extract`.

## Two switches in series

An operator binds a provider, from **Settings → AI → Web search** or the new `WEB_SEARCH_*` block, which makes the toggle *offerable*. A reader then ticks **Search the web** for a conversation. Unbound, nothing is declared to any model and both surfaces behave exactly as they did before.

The flag rides each request rather than the session — no migration, and a reload starts from off. Every call is billed, so off is the direction worth forgetting in. One answer makes at most ten calls (`maxWebCalls`), on either surface.

Plain **Search** mode is untouched: one round against the archive rendering a list of cards, with nowhere to put a web result. The server ignores the flag there, so a metered provider is not billed for it.

**What leaves the instance is a query the model wrote**, not the question the reader typed — and the model has the document text in front of it when it writes one. Assume a search can carry a phrase out of a document rather than only the subject of the question. Fetching runs on the provider's side, so the pages read see their address rather than yours.

## The parts worth reviewing

**Fetching is Tavily's job too.** `/extract` runs on their network, so a URL a model invents is never dialled from this process: no SSRF guard, no HTML-to-markdown code, one key. `websearch/tavily.go` follows `internal/ocr/mistral.go` — bounded response read (the base URL is admin-configurable), no retries, the provider's own error envelope decoded for the message.

**Ask AI had no tool loop at all.** `Chat()` was a single blocking completion. It now gathers over four rounds and then answers. The answer turn declares **no tools at all**, the way `answerResearch` does, rather than keeping them with `tool_choice: none` — that suppresses nothing for a model which emits its calls as message content, and the shipped default model does, so a schema still in front of it can end the turn in markup after the round has been billed. Passing `nil` takes the old path unchanged.

**`tavily` is the first SDK that serves neither a model nor a document**, which makes `CanOCR` the trap: it is a deny-list defaulting to `true`, so without an explicit arm a Tavily row binds happily to OCR and every uploaded document then fails at OCR time. Its frontend twin `providerServesPurpose` has the same shape and the same fix. `TestSDKListsMatchTheirPredicates` also asserted every valid SDK serves *some* binding, which a pure web-search SDK fails by design; it now counts `CanWebSearch`.

`bindWebSearch` clears on an empty id, the contract `bindEmbedding` has: `Apply` runs on every boot under `AI_MANAGED=1`, and a binding left standing when `WEB_SEARCH_*` is unset would be a one-way door, since the Settings page refuses that field in managed mode.

Deliberately **not** added: `PurposeWebSearch`, a `ServesPurpose` arm, a `needsModel` arm, an `Overrides` field. Those exist for per-request binding overrides, and this binding has no model and no override — `Resolve` is never called for it.

**Migration `1730000035`** widens `ai_providers.sdk` and adds `websearch_provider_id`. Without it a tavily row fails PocketBase's own select validation on any instance past its first boot, with nothing actionable in the message.

## Known limitation

Ask AI is a plain JSON POST, not SSE, so there are no progress steps while it searches — the user waits, then sees the answer with its citations. Converting that endpoint to SSE is the fix if it matters.

## Verification

`./scripts/test-all.sh` green: unit, API e2e, browser e2e (165), frontend unit, docs build. `docker build` clean.

Overlay branch: [`lemmary-dev#worktree-web-search-tool`](https://github.com/lemmary/lemmary-dev/tree/worktree-web-search-tool) — same name, pushed first. It adds `Options.WebSearch` (off by default, so the default pass keeps exercising the path where no tool is declared), a Tavily mock, and suites covering both switches, search mode ignoring the flag, document chat's new loop, the OCR-binding refusal that would otherwise fail silently, and `Apply` clearing the binding when the environment drops.


buldezir:
Both High findings were real and are fixed; the Medium copy issue too. Verified each against the code before changing anything.

**1. Ask AI losing a paid-for turn — confirmed, fixed.** `chatWithWeb` copied Search's fixed-round loop but not the shape that makes it terminate. Keeping tools declared with `tool_choice: none` suppresses nothing for a model that emits calls as message content, and that is the shipped default model. The round ran the tools, billed the provider, fell out of the loop and returned `chat did not produce an answer`.

Rather than porting `forceFinalAnswer`, the answer turn now declares **no tools at all** — the way `answerResearch` already does. `forceFinalAnswer` only exists because Search keeps the schema in front of the model; with nothing declared there is nothing to answer with but prose. The final round also no longer runs tools, so the last round cannot bill anything.

`TestChatAnswersEvenWhenTheLastRoundIsStillToolMarkup` fails on the old code with exactly the reported error — checked by stashing the fix and running it.

**2. Managed mode cannot unbind — confirmed, fixed.** `Apply` runs on every boot under `AI_MANAGED=1`, so the missing write was a one-way door. `bindWebSearch` now clears on an empty id, which is the contract `bindEmbedding` already had and which I said I was mirroring in the plan but did not.

**3. Cost and privacy copy — confirmed, fixed.** `maxChatToolRounds` is rounds, not calls, and both surfaces share the ten-call `maxWebCalls` budget; "at most four times per question" was wrong. The privacy claim was worse than imprecise — the query is written by the model, which has the document text in front of it, so a search can carry a phrase out of a document. Both corrected in `.env.example` and `docs/ai_providers.md`.

**Tests.** All four gaps covered. The research-loop tests read the real tool list off a captured request via the existing `toolNames` helper rather than re-deriving it, so they fail if `research.go` stops appending the tools — a hand-rolled mirror would have passed either way. Also added: the web result reaching the answer and surviving `validateCitations`, the emitted steps, and the per-run budget.

The `Apply`-level test lives in the overlay rather than beside `bindWebSearch`: reaching the clear through `Apply` upserts a provider, which needs a database.

**Nits.** Both fixed — `HasWebSearch` had landed between `HasEmbedding`'s doc comment and the function, and the provider-delete 409 did not mention web search though `ReferencedBySettings` now returns true for it.

`./scripts/test-all.sh` green. Overlay updated on the matching branch.

PR #84, opened 2026-09-15
Let a research chat branch into a fork, and say the web toggle is per turn

buldezir:
A Deep Search chat is one line of questions: its mode and binding are fixed for its lifetime and every follow-up appends to the same transcript, so trying a second line of enquiry meant either polluting the chat worth keeping or starting over without the context the answer depended on. **Research in fork**, beside Send, asks the question in a copy of the conversation instead and leaves the original exactly as it was.

## How it works

The fork rides the turn request rather than getting an endpoint of its own: `searchRequest` gains `fork_from`, read only when `session_id` is empty, as mode and the binding already are. Everything between loading the session and creating one in `prepareSearchTurn` already reads the source record and stays right for a fork — the mode-conflict 409 refuses forking a Search chat under `mode=research` for free, the binding comes off the conversation rather than the request, and `PriorHits` carries its evidence forward — so only the session the turn lands in changes. One round trip and one paid run; a separate `POST /chats/{id}/fork` would have made the client fork, navigate, wait for the load and then submit, three states to get wrong for nothing.

`chat.ForkSession` copies kind, document, mode, provider and model because a session refuses a turn that contradicts them, and the transcript with its hits and steps, renumbered from 1. `run_id` is dropped: it correlates a stored turn with the live request that produced it, and the copy produced none.

`discardOpenedSession` is the other half. A fork is never empty, so `DiscardEmptySession` would have left one behind after every failed run; the forked branch deletes outright.

## On the page

`forkRef` is a ref and not state because `submit` reads its send closure through a ref refreshed in an effect, so a `setState` in the click handler would not reach the send that click triggers. `onSessionSettled` now navigates whenever the turn landed somewhere other than where it was typed, keeping `replace` for the `created` case alone: Back should lead out of a fork and into the chat it branched from. Titles stack their mark rather than nesting a prefix, so a fork of a fork reads `⑂⑂ title` and says at a glance how far from the original it is. The composer row wraps and the textarea keeps a floor, because a page can now put two buttons in it.

## Unrelated, one sentence

The web toggle was already per request — the flag rides each turn and a reload starts from off — but its help text said only what it costs, so ticking it read as arming the conversation. It now says it applies to this question only, in the shared component, which covers Ask AI too.

## Tests

`ForkTitle` and `ForkSession` units here; the API and browser suites are in lemmary-dev on the matching branch (`worktree-fork-research`): the copy and the untouched source, `session_id` beating `fork_from`, another account's chat as a 404, and the button's flow end to end.

`./scripts/test-all.sh` passes.


buldezir:
Pushed `45a3d129`: the page now moves into the fork the moment the run opens it, rather than when the answer lands.

The server already named the copy in the stream's first frame — it is what makes a dropped connection recoverable — so the page follows that frame: URL, selected rail row and the research steps all belong to the fork from the start of the run.

The move is a promotion, not a switch. `useChatSession` keys its load effect on an ownership ref, and an unclaimed id reads as the user changing conversations: it would reset the transcript, drop the reply being watched and reload the fork before its turn is stored. The hook gains `adoptSession`, which claims the id before the URL follows so the effect takes its existing promotion no-op.

That makes `discardOpenedSession` wrong, so it is gone and the call sites return to `discardEmptySession`, which never deleted a fork anyway — a copy is never empty. A fork abandoned by a failed or cancelled run now survives: by then the user is standing in it with their question back in the composer, and deleting a conversation out from under someone looking at it is worse than leaving one they asked for and can delete.

Overlay branch updated to `44d8bcd`: the stream's first frame must name the copy and the turn must be stored in it, and the browser spec now asserts the rail row is `aria-current` and the in-flight question survives the move. Neither stalls a provider, so they pin the contract and the outcome rather than the instant.

PR #90, opened 2026-09-15
Keep a research chat's thinking out of its transcript, and move into it on create

buldezir:
Three things a research chat was getting wrong.

## The model's thinking was rendered as an answer

An assistant row that carries tool calls is machinery, but `chat.Visible` only
tested the role and whether the content was empty. So the narration a model
writes on its way to a search — "Let me search the archive for that." — came
back from the server as an answer bubble, contradicted moments later by the
real answer.

`Visible` now takes whether the row made calls, with a `VisibleRecord` helper
for the four callers holding a `*core.Record`. Fixing it at the shared
predicate rather than in the transcript fold clears two siblings of the same
bug:

- `message_count` counted narration as a message, so the rail's count drifted
  from what the chat shows.
- `Unfinished()` read a run that died *between* a narrated tool call and its
  result as finished, hiding the **Continue** button that exists to pick that
  work back up. (A run that died before the model said anything was already
  handled — the empty-content case.)

The narration folds into the research-steps trail instead, which is what the
comments around this code already claimed happened.

## A new research chat now moves into its URL on create

The page moved to `/rag/research/{id}` only once the answer arrived, but the
session exists from the first frame of the stream. A reload during a long run
therefore landed on the empty page the question was typed on, with the run
still going and nowhere to watch it. It now adopts and navigates on the
`session` frame, the same way a fork already did, with `created: true` so Back
skips the orphaned empty page.

Search deliberately stays put: a search turn that fails discards the session it
opened, and the URL would be left pointing at a chat that no longer exists.

One visible side effect: a research chat takes its rail row as soon as the run
starts rather than when it finishes.

## `AI_MODEL_CATALOG_URL` is documented as it behaves

`getEnv` reads empty as unset, so `AI_MODEL_CATALOG_URL=` falls back to pi.dev
rather than turning the lookup off. `.env.example`, `docs/ai_providers.md` and
both `aienv.go` comments said the opposite. The off switch is the per-provider
**None** catalogue in Settings.

The `Catalog` disabled path for a blank base URL is real but unreachable from
the environment. Docs now say so; making the old wording true instead would be
a two-line `os.LookupEnv` change, if that is preferred.

## Verification

`./scripts/test-all.sh` passes against lemmary/lemmary-dev#31, which must merge
alongside this. Both branches are named `worktree-research-improvements`, so
the PR job picks up the matching suite rather than falling back to `main`.


buldezir:
Both findings check out. Fixed in 3e833d8d, with the coverage in lemmary/lemmary-dev#31.

**1 — the failure path.** Confirmed, and worse than stated in one way: `submit`'s catch also leaves `unfinished` false, and the *notice it would have shown is inert anyway* — `continueTurn()` calls `chat.submit()`, which returns early on an empty composer, and a reloaded chat always has one. So Continue has never worked from a reload; adopting on create just made that the common path.

A run that claimed a conversation mid-flight has its question and its tool work on disk, so a failure there is an unfinished turn, not a lost send: the bubble stays, the composer stays empty, `unfinished` is set, and the way on is Continue — the same state a reload of that chat already produced. `submit({ resume: true })` waives the text it does not send (the server ignores `content` when `resume` is set).

Worth noting the same adopt-then-fail path predates this PR on forks, and is fixed with it. Search never adopts — no fork send of its own, and it discards the session a failed turn opened, which is why it deliberately still stays put.

**2 — DSML.** Confirmed; `ContentHasDSMLToolCalls` is exported and folded into `Visible`, so both routes to a tool round read the same. Store test added for a DSML row: not visible, `Unfinished` true, `message_count` 1.

**On coverage.** No client test existed because the frontend suite is pure `src/lib` modules — no jsdom, no testing-library — and adding both for one branch was the wrong trade. It went to the overlay instead, which needed a new affordance: `FailCompletions` is a Go harness call the browser suite cannot reach, so the mock now refuses the first completion carrying a `refuse-once-<id>` marker and answers every later one. The spec asserts the whole arc — failure inside the adopted conversation, question kept, composer empty, Continue offered, same state after reload, then continuing to an answer with no second copy of the question. I checked it fails with either half of the fix reverted.

PR #92, opened 2026-09-16
Anthropic SDK provider

buldezir:
Anthropic was already on the wire here — `anthropic-sdk-go` has been a direct dependency since `opencode` got its own SDK, because a third of that gateway's catalogue is served on `/v1/messages`, and `internal/opencode/messages.go` is a complete, tested translation between `openai.ChatCompletionNewParams` and `anthropic.MessageNewParams`. What was missing was a way for an operator holding an Anthropic key to use any of it.

`AI_SDK=anthropic`, or an **Anthropic** row in Settings, now binds `api.anthropic.com` for extraction, chat, Deep Research and OCR. Not embeddings: that host serves none at all, so `CanEmbed` refuses it and Deep Search's vectors still want a second provider — the same shape as `opencode` and `chatgpt`.

## The translation moved

`internal/opencode` is documented as the OpenCode Go SDK and had no business being imported to reach Anthropic. The translation is now a neutral leaf package, `internal/messages`; `opencode` keeps its endpoint routing table and nothing else. The files moved with `git mv`, so the history is intact.

The `sdk` travels with every call there, which is what gates the only two per-SDK differences:

- `x-opencode-session` — OpenCode's routing key, meaningless to Anthropic, which would only see an unknown header carrying a conversation id on every request.
- `output_config.effort` — Anthropic's, and the MiniMax and Qwen models behind OpenCode's `/messages` do not take it.

## Effort defaults to low

Claude's own default is `high`. This codebase's completions are extraction, search and distillation over documents already in hand, paid for by the token, so requests on this SDK ask for `low` unless a caller says otherwise.

The default lives in `internal/ai` rather than in the translation because it has to be droppable: `output_config.effort` is only understood by Claude 4.5 and later. A refusal is retried once without the field and remembered per model in `modelnotes.go`, beside the two degradations `completeChat` already carries — so a bound older model costs one retry, not every request.

## Two gaps a first-class provider made visible

- `ListModels` signs Anthropic with `x-api-key` and `anthropic-version` rather than a bearer (a request carrying both is read as OAuth and refused), and reads `display_name` and `max_input_tokens`.
- `ProviderErrorMessage` learns `*anthropic.Error`, which until now fell through to the generic branch and lost its status code — a gap the opencode `/messages` path already had.

## Verification

`./scripts/test-all.sh` passes every stage — unit, API e2e, browser e2e, frontend unit, frontend build, docs build — and `docker build -t lemmary:local .` succeeds. `go.mod` is unchanged.

No overlay branch: the suites pin specific SDKs rather than asserting the full list, so nothing there needed updating and the PR job's fallback to overlay `main` is what this was verified against.

New tests: an `anthropic` row in the `TestSDKCapabilities` and `TestSDKListsMatchTheirPredicates` tables; the effort mapping and the session-header gate per SDK in `internal/messages`; a full round trip, the low-effort default, the retry-and-remember, and streaming in `internal/ai/anthropic_test.go`; the models-listing headers; the error rendering; `AI_SDK=anthropic` seeding and the embeddings refusal; the OCR factory; and the migration up/down.


buldezir:
Merged `main` (the migration renumbered to `1730000040`; `1730000039` went to tag colours) and worked the review. Findings 1, 2, 3 and 5 were real; 4 I've left alone, with reasons below.

**1 — temperature, and 5 — the effort matcher.** Both confirmed. Every completion here sets a temperature through `CompletionTemperature`, whose only exemption was gpt-5/o-series, and Claude removed the parameter after Opus 4.6 — so `claude-opus-5` would have 400'd on the first extraction. The effort matcher fired on any 400 mentioning the word, and wrote its note before the retry had worked.

**2 — thinking blocks.** Confirmed, and the reason it cannot be fixed by replaying them: `internal/messages` converts every reply to an `openai.ChatCompletion`, which has nowhere to keep a thinking block, so a tool loop rebuilding the assistant turn cannot carry one. Thinking is now off, with a `ponytail:` comment naming what carrying it would cost if answers suffer.

Rather than three bolt-ons, the three fields are one ladder in `internal/ai/messages.go`: sent hopefully, dropped one per retry when a 400 names them, remembered per **model and field** only once a request without them has succeeded. Deliberately no model-version table — the catalogue reports none of this and a table would be stale by the next release. The two fields with no OpenAI counterpart moved into a `messages.Options` the caller owns, which is what makes them droppable.

**3 — dual auth.** Fixed: the bearer is OpenCode's alone on completions, as it already was on the model list.

**4 — `AI_MODEL` default.** Confirmed but left as is: `parseLLM` falls back to `DefaultExtractModel` for `mistral` and `openrouter` too, so it is pre-existing behaviour across every non-OpenAI SDK rather than something this PR introduces; a per-SDK default belongs in its own change. The test comment that called it a complete configuration was overclaiming and now says the opposite.

**Tests.** `providers.test.ts` names `anthropic` in the capability and key tables and pins its default base URL. New Go coverage: temperature refused and retried, disabled-thinking refused and retried (the Fable case), a refusal about the account *not* retried and nothing remembered, the streaming retry, absence of the bearer, and the `Options` → `output_config`/`thinking` mapping per SDK.

`./scripts/test-all.sh` passes in full. Worth noting: the first run after the merge failed two `search-sessions.spec.ts` browser tests, which turned out to be a stale overlay — `lemmary-dev` had its own counterpart commit — not these changes.

PR #93, opened 2026-09-16
Upgrade openai-go from v1.12.0 to v3.61.0

buldezir:
Two major versions and a module-path change behind. What prompted this was a question about OpenAI's [migrate-to-responses incremental rollout checklist](https://developers.openai.com/api/docs/guides/migrate-to-responses#incremental-rollout-checklist) — we already satisfy it, and the stale pin was the one real finding.

### Why the checklist needs nothing

`/responses` is already spoken wherever a model requires it, `store: false` is set on both Responses paths, structured outputs go through `text.format`, streaming reads typed events, and the chat-shaped internal currency is what keeps the six non-OpenAI SDKs working. Chat Completions has no announced sunset. The one unmet item, replaying encrypted reasoning items, is deferred: a reasoning item has no home in a chat-shaped thread, so it needs a carrier through `ThreadMessage` and both translators rather than a patch. `ResponseNewParams.Include` and `.Reasoning` both exist in v3, so nothing here blocks it.

### What actually broke

Five symbols, all from the v2.0.0 rename that made room for custom tools:

| Removed | Replacement |
| --- | --- |
| `ChatCompletionToolParam{Function: f}` | `openai.ChatCompletionFunctionTool(f)` |
| `[]ChatCompletionToolParam` | `[]ChatCompletionToolUnionParam` |
| `ChatCompletionMessageToolCall` | `ChatCompletionMessageToolCallUnion` — flattened, so `Type` is a plain string now |
| `ChatCompletionMessageToolCallFunction` | `ChatCompletionMessageFunctionToolCallFunction` |
| `ChatCompletionMessageToolCallParam` | `ChatCompletionMessageToolCallUnionParam{OfFunction: …}` |

Sites reading a tool or tool call off a union go through `GetFunction()`/`OfFunction` and skip a non-function variant — impossible today, a nil dereference if it ever happened. Two further v3 changes the compiler names: the Responses stream event `Delta` is a plain `string`, and an output item's `Arguments` became a union read via `.OfString`.

### The one silent break

`ResponseInputItemParamOfFunctionCallOutput` no longer takes the call id — its output field became a union and the signature lost the first parameter. Dropping that argument to satisfy the compiler sends a function call output correlated to nothing, breaking every tool loop on the Responses path. `CallID` is set on the item explicitly instead. `responses_test.go:346` already asserted `call_id` on the translated item, so the regression was covered.

Also collapses a comment in `completeChat` that a bad paste had left duplicated and truncated mid-sentence.

Out of scope: `anthropic-sdk-go` stays at v1.71.0.

### Verification

`./scripts/test-all.sh` — all seven stages green (unit 109s, API e2e 90s, browser e2e 34s, frontend unit, frontend build, e2eserver build, docs). `docker build -t lemmary:local .` succeeds. No overlay branch: the change needs no e2e update, so the PR job falls back to `main`, which is what the local run used.


buldezir:
Good catch — confirmed and fixed in 58a98c62.

Verified the mechanism against the v3.61 source rather than taking it on trust. `chatcompletion_helpers.go` switches on `v.AsAny()`, and an unclassified call is still *appended* as a zero `ChatCompletionMessageToolCallUnionParam`, which `MarshalUnion` renders as `null`. v1.12.0 copied `ID`/`Function` across unconditionally. Empirically:

```
wire without "type" -> Type="" ID="call_9" Name="search_documents"
replayed            -> {"tool_calls":[null],"role":"assistant"}
```

You were also right that the `OfFunction == nil` guards made it quiet rather than loud: a dropped call plus an orphaned `tool` result instead of an error.

The fill-in sits at the one choke point instead of the three tool loops — `Complete` now wraps what it used to do, so `completeChat`'s degradation and retry paths are covered too, not just the three `ToParam()` sites. `chatCompletionFrom` in both translators already sets the type, so nothing there changes.

Both tests you asked for:

- `TestAToolCallWithoutATypeIsStillReplayable` — mutation-checked. With the fill-in disabled it fails on exactly `{"tool_calls":[null],"role":"assistant"}`.
- `TestAResponsesToolCallSurvivesBeingReplayed` — closes the gap that nothing fed a translated response back through `responsesInputFrom`: a Responses `function_call` comes back as `function_call` + `function_call_output` with `call_id` intact through both translations.

`./scripts/test-all.sh` green again (all seven stages, 101s wall clock).

PR #99, opened 2026-09-17
Settings: two LLM bindings, General AI and an optional Advanced model

buldezir:
## Why

The AI settings page offered one language model in "Simple setup" and four in "Advanced setup" (Extraction, Chat, Search, Deep Search helper). Nobody needs four knobs. Two cover every real case: one model for everything, and optionally a stronger one for the Deep Research reasoning loop, which is the only place where a few expensive calls decide the answer.

## What

- **General AI** is the existing `extract_*` binding (name kept, no API break for extraction). It now serves extraction, split detection, Ask AI, AI assisted search, Deep Research's bulk document reads, and AI tag assignment.
- **Advanced** is a new optional `research_*` binding. It drives only the Deep Research reasoning loop; empty means General AI does that too. The fallback lives in `Config.ResearchBinding()` rather than in the stored fields, so `GET /settings` reports an empty research binding as empty.
- `chat_*`, `search_*` and `search_helper_*` are dropped by migration `1730000041`. A search binding that differed from extraction is carried into `research_*`; the helper is not, since the general model does that work now.
- `AI_SEARCH_HELPER_MODEL` becomes `AI_RESEARCH_MODEL`, documented in `.env.example` with what it costs. Managed mode re-applies it every boot; unset clears it.
- The search endpoint picks the default pair by mode (research binding for a research chat, general for search). The pickable-providers endpoint answers `binding=research` with the research pair; the research page asks for it, so its composer names the model that will actually answer.
- The Simple/Advanced toggle and the four per-job fields are gone from the settings page. The setup wizard patches only `extract_*`. Docs and README follow.
- The per-request `Overrides{Chat, Search, Extract, ...}` struct is unchanged: it names the client being rebuilt for one request, not a stored binding.

## Verification

`./scripts/test-all.sh`: unit, API e2e, frontend unit, build and browser e2e all pass. `docker build -t lemmary:local .` passes.

Overlay branch `worktree-model-selection-refactor` in lemmary-dev (pushed) updates the API and browser suites and adds a case: research opens on the Advanced model when bound and on the general one when not, while plain search stays on the general model.

Screenshots under `docs/screenshots` that show the old Models section are stale and not regenerated here.


buldezir:
Addressed in 9a9fa440 (public) and 9d34a35 (overlay):

- **Unrecorded research binding ran on General AI** — fixed. The stored-or-requested binding is now resolved against the mode's configured pair before the snapshot is built, so a conversation that stored nothing follows Settings (Advanced for research, General for search) and agrees with `contextWindowFor`. New API e2e blanks a research session's binding, binds an Advanced model, and asserts the follow-up's provider request names it.
- **Migration carried a half pair** — fixed. Only a search binding with both halves set (and differing from extraction) is carried; a provider-without-model row leaves `research_*` empty. Test case added. `FindRecordById` now fails the migration on anything but no-rows.
- **Model without provider accepted** — fixed. `applySettingsPatch` refuses either half alone; the unit test asserts both directions.
- **Stale `Snapshot.SearchHelper` comment** — fixed.
- **Distinct chat binding not carried** — intentional. Chat customizations were the "use a different model for Ask AI" case, and Ask AI now runs on General AI by design; there is no field to carry them into.
- **`conversationSnapshot` falling back to General when a default research override fails to resolve** — left as is. Deleting a bound provider is refused via `ReferencedBySettings`, so the pair only fails to resolve when the row is gone by other means, and the pre-existing behaviour (warn, continue on the configured snapshot) is kept.

Full suite green: unit, API e2e, frontend unit, build, browser e2e.

